PR #4, opened 2019-03-14
AG-110 - Support custom transaction isolation levels

barreiro:


barreiro:
@jesperpedersen I though about that, but it breaks backwards compatibility.

jesperpedersen:
Sure, go ahead and merge as is

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionFactoryConfiguration.java
```diff
@@ -21,7 +21,7 @@
 
     Class<?> connectionProviderClass();
 
-    TransactionIsolation jdbcTransactionIsolation();
+    IsolationLevel jdbcTransactionIsolation();
```
dmlloyd: This breaks binary compatibility, if that matters.

agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalConnectionFactoryConfigurationSupplier.java
```diff
@@ -98,12 +99,18 @@ public AgroalConnectionFactoryConfigurationSupplier connectionProviderClassName(
         }
     }
 
-    public AgroalConnectionFactoryConfigurationSupplier jdbcTransactionIsolation(TransactionIsolation transactionIsolationLevel) {
+    public AgroalConnectionFactoryConfigurationSupplier jdbcTransactionIsolation(IsolationLevel transactionIsolationLevel) {
```
dmlloyd: This also breaks binary compatibility; you could add an overload though.

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionFactoryConfiguration.java
barreiro: I don't think it matters. This interface is not intended to be called/implemented outside agroal.

agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalConnectionFactoryConfigurationSupplier.java
barreiro: On a second though, yes, this method does not need to be relaxed.

PR #5, opened 2019-03-15
AG-111 - Connection lifetime feature

barreiro:


jesperpedersen:
Describe the scenarios where this code is executed.

barreiro:
@jesperpedersen it's in the issue:

> The reason behind this feature is firewalls with strict connection timeouts (even if the connection is active). In order to prevent those connections from becoming invalid, they should be removed before the firewall timeout.

jesperpedersen:
Flushing active connections is an area I don't think we should explore. 

If there are deployment requirements around timeouts from firewalls then the application needs to keep track of it, and call `Statement.cancel()` before the timeout happens, deal with `SQLException` after and so on.

gsmet:
I haven't checked the implementation but the idea is to not provide connections that are older than a certain age.

It's implemented in a lot of pools and I had to use it several times on customers site.

My preference on the name would be `maxLifetime` instead of `lifetime`.

barreiro:
Renamed to ```maxLifetime```

PR #15, opened 2019-10-23
"Info" logging for initial pool size

Sanne:
In Quarkus we expect the "initial size" of the pool to be set by default (intentionally) as lower than the min-size; but when doing so this produces a concerning warning to end users.

Would you agree to "demote" the severity of this particular message?

I've introduced an "info" level for logging in the first commit, then changed that particular message to use it.

Let me know if you have better ideas! Thanks


jesperpedersen:
I don't think this is the correct fix.

If min-size is >= 0, and initial-size isn't defined then initial-size should be min-size

Sanne:
hi @jesperpedersen !
for cloud deployments it's desireable to allow the users to configure a "low rampup", including zero, as a different starting point than what the min-size should be during sustained load.
I have a requirement in Quarkus to allow this; indeed I'm not persuaded about how we patched this here, I had also suggested an alternative which is to just fire an event like "logger.warnInitialSizeOutOfBounds( min, max, initial )" - that would allow platforms like Q to do weird stuff without having an impact on other potential users of Agroal - but Luis felt my other proposal was over-engineered ..

either way, I agree we should polish this somehow. Just please allow our requirement as well.

jesperpedersen:
In the case, `min-size > 0 && initial-size == 0`, there shouldn't be a warning/info as it is a valid configuration. That is if `initial-size` is explicit set.

Sanne:
yes that sounds good to me. Before this last release we were getting a WARN, I just needed to get rid of that.

PR #16, opened 2019-10-25
AG-131 - Change data structure used to track JDBC resources

barreiro:
The current data structure is optimized for when the number of reads far exceeds the number of writes, which is the opposite of what we want here. 

I believe that because there will be mostly add / remove operations, a data structure backed by an array may not be the best fit for this use case, therefore ended up going with ``ConcurrentLinkedQueue``.


Sanne:
Change looks good!

I was hoping to benchmark it too, but that's going to take a while longer as all my current figures were generated while having `quarkus.datasource.detect-statement-leaks=false`.

barreiro:
Same here. The benchmark that I have does not create any Statements, so it only shows difference in time and size to create the list, doesn't tell anything about modifications.

PR #17, opened 2020-03-31
AG-136 - AgroalPoolInterceptor

barreiro:


jesperpedersen:
ISE -> AgroalException would be better

barreiro:
I'm leaving IAE for now as there isn't a custom exception type in Agroal.

jesperpedersen:
Ok

PR #23, opened 2020-05-26
AG-143 - Add FlushMode.LEAK

barreiro:


agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java
```diff
@@ -185,12 +187,16 @@ public boolean isActive() {
...
     }
 
-    public long getLastAccess() {
-        return lastAccess;
+    public void touch() {
+        lastAccess = nanoTime();
+    }
+
+    public boolean isLeak(Duration timeout) {
+        return isActive() && !enlisted && isIdle( timeout );
```
barreiro, line 195: I'm assuming that if a connection is enlisted it's never to be considered a leak, as the transaction will eventually complete.

jesperpedersen: Correct

PR #25, opened 2020-09-08
AG-147 - Spring module initial commit

barreiro:


agroal-springframework/src/main/java/io/agroal/springframework/AgroalDataSourceConfiguration.java
```diff
@@ -0,0 +1,40 @@
+// Copyright (C) 2020 Red Hat, Inc. and individual contributors as indicated by the @author tags.
+// You may not use this file except in compliance with the Apache License, Version 2.0.
+
+package io.agroal.springframework;
```
geoand, line 4: As this is spring boot specific, maybe the package name should reflect that?

agroal-springframework/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <artifactId>spring-boot-autoconfigure</artifactId>
+            <version>${version.org.springframework.boot}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-jpa</artifactId>
+            <version>${version.org.springframework.boot}</version>
+            <scope>provided</scope>
+        </dependency>
```
geoand, line 41: Are you sure this dependency is needed?

agroal-springframework/src/main/java/io/agroal/springframework/AgroalDataSourceConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.transaction.jta.JtaTransactionManager;
+
+import javax.sql.DataSource;
+
+/**
+ * @author <a href="lbarreiro@redhat.com">Luis Barreiro</a>
+ */
+@Configuration
```
geoand, line 21: You might want to use:

```
@Configuration(proxyBeanMethods = false)
``` 

Which is the norm these days

geoand (approved):
LGTM, but I would like @gytis to take a look as well as he is the SB+Narayana SME

agroal-springframework/pom.xml
gytis: Seems that it should be enough to have `spring-tx` no?

barreiro: Changed dependency to `spring-tx`. Updated `narayana-spring-boot-starter` to `2.4.0` and added test case to check the transaction integration is engaged. Thanks @gytis for the quick fix!

agroal-springframework/src/main/java/io/agroal/springframework/AgroalDataSourceConfiguration.java
barreiro: Changed module name to `agroal-spring-boot-starter` and package name to `io.agroal.springframework.boot`

agroal-springframework/src/main/java/io/agroal/springframework/AgroalDataSourceConfiguration.java
barreiro: Good suggestion. Applied.

PR #27, opened 2021-02-22
AG-155 - Deadlock warning on multiple getConnection() by the same Thread

barreiro:


jesperpedersen:
Hmm, I don't know about this...

Non-transacted cases can obtain multiple connections which needs to be supported. If this happens in a transacted case then something more fundamental is wrong and we def need a trace of the call stack.

I think there needs to a more dedicated property for this that checks for this condition - it could be part of leak detection too though.

Try and fold this work into a property that verifies transaction semantics vs physical connections - or a more enhanced version of leak detection.

Also, it is starvation - not deadlock.

jesperpedersen:
I still think this needs more work - it should depend on the functionality added in #28 in order to give a more precise warning in the various cases.

barreiro:
superseded by #31

PR #30, opened 2021-03-03
AG-158 - Add error codes to XAException

barreiro:


agroal-narayana/src/main/java/io/agroal/narayana/LocalXAResource.java
```diff
@@ -36,64 +40,64 @@ public Object getConnection() throws Throwable {
...
+            throw xaException( XAER_NOTA, "Invalid xid to transactionCommit" );
         }
 
         currentXid = null;
         try {
             transactionAware.transactionCommit();
         } catch ( Exception t ) {
             transactionAware.setFlushOnly();
-            throw new XAException( "Error trying to transactionCommit local transaction: " + t.getMessage() );
+            throw xaException( onePhase ? XA_RBROLLBACK : XAER_RMFAIL, "Error trying to transactionCommit local transaction: " + t.getMessage(), t );
```
barreiro, line 70: XAER_RMFAIL leads to HeuristicRollbackException, while XA_RBROLLBACK leads to RollbackException. I believe the former makes little sense when committing a single resource, therefore opted for the latter when in single phase.

jesperpedersen:
Yeah, likely the best way to resolve that.

@mmusgrov - Could you take a quick look; main driver is 1PC in JDBC scenario

ochaloup (commented):
Just adding my observations. @mmusgrov may correct me.

The XA spec (https://pubs.opengroup.org/onlinepubs/009680699/toc.pdf) expects for commit/rollback the `XAException` `XAER_RMFAIL` error code as the RM is currently unavailable and thus it may be retried. But for non-XA there won't be a chance for recovery manager to retry as there was no prepare phase and the failure on connection means failure of the commit/rollback call.

barreiro:
Thanks for your review @ochaloup. I changed `XAER_RMFAIL` to other error codes, namely `XAER_RMERR`, based on the observation that recovery is not possible.

Please note that part of the problem is that the JDBC driver throwing an `SQLException` is vague information, as there may be several causes for it and there is no easy way to tell (in a standard way) the real reason. It's not possible to distinguish between a temporary failure (like a communication failure) and a permanent one (for example, the DB being in a state that can't commit). For now all failures are to be treated as permanent.

PR #31, opened 2021-03-03
AG-155 - Add setting for multiple getConnection() by the same Thread

barreiro:
An alternative to #27 using a setting


jesperpedersen:
Yeah, this is better.

We could go with this first, and add an unset value check later.

PR #36, opened 2021-06-22
AG-173 - Make local cache customizable

barreiro:


jesperpedersen:
I would have the `Acquirable` and a `ConnectionCache` interface as the main interfaces for this feature.

Then have `LocalConnectionCache` extend `ConnectionCache` (or implement) which people can use. Or supply their own implementation. Mixing it - as it is now - is a bit confusing for external providers.

Where does `.fixed(4)` come from ? And, why is `.fixed()` needed ? Apps would expect `.single()` or `.full()`, no ?

barreiro:
> Where does `.fixed(4)` come from ? And, why is `.fixed()` needed ? Apps would expect `.single()` or `.full()`, no ?

`fixed(4)` is the current implementation. will change this default after this commit is merged.

jesperpedersen:
Good to go, as you said you wanted to fix the `.fixed()` case after

jesperpedersen (approved):

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
```diff
@@ -0,0 +1,137 @@
...
+                            return element;
+                        }
+                    }
+                }
+                return null;
+            }
+
+            @Override
+            public void put(Acquirable acquirable) {
+                Acquirable[] cacheArray = threadLocal.get();
```
stuartwdouglas, line 79: Why does the one above have an isAcquirable() check but not this one?

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
```diff
@@ -0,0 +1,137 @@
...
+                int last = cacheArray.length - 1;
+                if ( cacheArray[last] != null ) { // replace last
```
stuartwdouglas, line 81: What is the reasoning behind the logic here? It fills the last spot first, then fills from the front, but aquires from the back?

If you have two items in here this means you get one at the front and one at the back? If there is a reason for this can you add a comment explaining why?

agroal-api/src/main/java/io/agroal/api/cache/ConnectionCache.java
```diff
@@ -0,0 +1,53 @@
...
+// You may not use this file except in compliance with the Apache License, Version 2.0.
+
+package io.agroal.api.cache;
+
+/**
+ * Interface for a cache of connections. It's intended for mapping connections to executing threads.
+ *
+ * @author <a href="lbarreiro@redhat.com">Luis Barreiro</a>
+ */
+public interface ConnectionCache {
```
stuartwdouglas, line 11: Shouldn't this have <T extends Acquirable> and get/put take type T so there is no need to cast?

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
```diff
@@ -0,0 +1,137 @@
...
+                    } else if ( acquirable.isAcquirable() ) {
+                        queue.addLast( acquirable );
+                    }
+                }
+                return null;
+            }
+
+            @Override
+            public void put(Acquirable acquirable) {
+                threadLocal.get().addFirst( acquirable );
```
stuartwdouglas, line 127: Shouldn't this have an isAcquirable check?

stuartwdouglas (changes requested):
This should be ok for our needs in Quarkus, but it looks like there are a few other minor issues

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
barreiro: yes, it's missing. Pool calls put on all connection return even when these are to be flushed.

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
barreiro: the logic is that the latest put has always preference, as it will be likely that can be acquired again. the other slots are remembered in the hope that whoever acquired it had released it between. 

anyway, this implementation is to mimic the existing behavior (for benchmarking purposes) and the plan is to change the default behavior (in favor of one of the other implementations, likely `single()`).

agroal-api/src/main/java/io/agroal/api/cache/ConnectionCache.java
barreiro: There will always be a need to cast, as the `ConnectionHandler` class is not on the API.

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
barreiro: yes, this needs to be fixed.

agroal-api/src/main/java/io/agroal/api/cache/LocalConnectionCache.java
barreiro: changed the insert order to fill from last to first.

PR #39, opened 2021-08-27
Log error if task fails

stuartwdouglas:


barreiro:
@stuartwdouglas superseded by #40

PR #44, opened 2021-09-17
AG-180 - Health check feature

barreiro:


agroal-hikari/src/main/java/io/agroal/hikari/HikariUnderTheCovers.java
```diff
@@ -119,6 +119,13 @@ public void flush(FlushMode mode) {
         hikari.getHikariPoolMXBean().softEvictConnections();
     }
 
+    @Override
+    public boolean isHealthy(boolean connect) throws SQLException {
+        try (Connection connection = getConnection() ) {
```
jesperpedersen, line 124: Check `connect`

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -448,12 +460,38 @@ public long awaitingCount() {
         return synchronizer.getQueueLength();
     }
 
+    // --- health check //
+
+    @Override
+    public boolean isHealthy(boolean connect) throws SQLException {
+        ConnectionHandler healthHandler;
```
jesperpedersen, line 467: If the pool is full, and all connections are checked out - this needs to work as well

agroal-pool/src/main/java/io/agroal/pool/Poolless.java
```diff
@@ -313,6 +317,31 @@ public long awaitingCount() {
         return synchronizer.getQueueLength();
     }
 
+    // --- health check //
+
+    @Override
+    public boolean isHealthy(boolean connect) throws SQLException {
+        // the main difference here is that connect will not block (and potentially timeout) on full pool
+        if ( connect ) {
```
jesperpedersen, line 325: You can't go over max-pool-size as that setting may be enforced by another layer or the database itself

jesperpedersen (changes requested):

barreiro:
Added warning that creating new connections can cause the pool size to go over max.

agroal-pool/src/main/java/io/agroal/pool/Poolless.java
barreiro: I can go over max! it's a matter for who dimensions the system.

That's why there are two modes and only one of them causes the size to go over max.

agroal-api/src/main/java/io/agroal/api/AgroalDataSource.java
```diff
@@ -81,6 +81,15 @@ static AgroalDataSource from(AgroalDataSourceConfiguration configuration, Agroal
      */
     List<AgroalPoolInterceptor> getPoolInterceptors();
 
+    /**
+     * Performs a health check. The connect parameter determines that a new database connection is established for this purpose, otherwise attempts to get a connection from the pool.
+     *
+     * WARNING: Using a new connection mau cause the size of the pool to go over max-size.
```
jesperpedersen, line 87: may

jesperpedersen (approved):

PR #47, opened 2021-12-23
Fix : Change ConditionalOnProperty havingValue io.agroal.springframework.AgroalDataSource by io.agroal.springframework.boot.AgroalDataSource

vincesi:
It must be io.agroal.springframework.boot.AgroalDataSource because the io.agroal.springframework.AgroalDataSource class does not exist.


barreiro:
merged in 250f3036aef1b5f08ddcba464718aeb2bbc99464

PR #48, opened 2022-02-22
Treat all postgres connection exceptions as fatal

stuartwdouglas:


agroal-api/src/main/java/io/agroal/api/exceptionsorter/PostgreSQLExceptionSorter.java
```diff
@@ -17,6 +17,6 @@
     @Override
     @SuppressWarnings( "CallToSuspiciousStringMethod" )
     public boolean isFatal(SQLException e) {
-        return "08006".equals( e.getSQLState() );
+        return e.getSQLState().startsWith("08"); //08 is connection exceptions
```
gastaldi, line 20: There are cases where SQLState may be null

```suggestion
	String sqlState = e.getSQLState();
        return sqlState != null && sqlState.startsWith("08"); //08 is connection exceptions
```

barreiro:
merged in e8bb47d44977fb7e00f2de21ccecbc4263d7f79a

PR #52, opened 2022-03-25
AG-191 - LoginTimeout feature

barreiro:


agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionFactoryConfiguration.java
```diff
@@ -27,6 +28,11 @@
      */
     boolean trackJdbcResources();
 
+    /**
+     * Maximum time to wait while attempting to connect to a database. Resolution in seconds.
+     */
+    Duration loginTimeout();
```
barreiro, line 34: Opted for Duration for consistency with other time based settings.

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -122,6 +122,7 @@
     /**
      * The maximum amount of time a thread may be blocked waiting for a connection. If this time expires and still no connection is available, an exception is thrown.
      * A duration of {@link Duration#ZERO} means that a thread will wait indefinitely.
+     * In Pool-less this timeout can add to {@link AgroalConnectionFactoryConfiguration#loginTimeout()}.
```
barreiro, line 125: In Pool-less, there is no background thread and therefore there can be a login timeout period after an (almost) acquisition timeout period. Added a note here and will update documentation as well.

barreiro:
@jesperpedersen the defaults for `DriverManager` and `javax.sql.DataSource` / `XADataSource` are `0`. I don't get what would be the advantage of loading the default from `DriverManager` (we don't set a value unless it was specified in the configuration).

jesperpedersen:
Did you verify that with a URL of `.../test?loginTimeout=2` ?

barreiro:
> Did you verify that with a URL of `.../test?loginTimeout=2` ?

No, because that is specific to each driver. But I checked `.jdbc.properties( "LoginTimeout", "10" )`.

PR #53, opened 2022-04-11
AG-192 Provide an agroal-narayana-jakarta artifact

gsmet:
Also test everything with the Jakarta artifact, excluding the Spring tests
though.
It makes use of the Jakarta transformer and an approach similar to what
we do for other projects.

I would appreciate if you could do a release soon as it's kinda in my way :).

Thanks!


Sanne:
Looks great - but I can't approve it myself :)

PR #54, opened 2022-04-22
AG-192 - Add Jakarta artifact

barreiro:


barreiro:
An alternative to #53

gsmet:
So, with your solution:
- you point to an incorrect `jboss-transaction-spi` dependency (it should use `jboss-transaction-spi-jakarta`)
- it's not tested at all

I fail to see how this is an improvement over what I did.

FWIW, the new artifact solution is not ideal but is used all over JBoss things because more practical. And the idea is to drop the artifact once we move to Jakarta fully (and release a 2.0 solely based on Jakarta packages).
And you could add a relocation to make sure people are not lost.

PR #57, opened 2022-10-14
AG-197 - Pool goes under min-size when connections go over max-lifetime

barreiro:


agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -581,16 +582,11 @@ private final class FlushTask implements Runnable {
 
         @Override
         public void run() {
-            if ( handler != null ) {
+            for ( ConnectionHandler handler : handler != null ? Collections.singleton( handler ) : allConnections ) {
```
jesperpedersen, line 585: I would do

```
Collection handlers = null;
if (handler != null) {
  handlers = Collections.singleton(handler);
} else {
  handlers = allConnections;
}
for (ConnectionHandler ch : handlers) {
```

Having `ConnectionHandler handler : ` as a variable shadow is confusing

barreiro: renamed the variable to remove shadowing

jesperpedersen (approved):

PR #64, opened 2023-03-30
Close connections during CRaC snapshotting

rvansa:
This change intends to support an application using Agroal (e.g. Quarkus app) to perform the Checkpoint and Restore on JVM implementing this, specifically using [OpenJDK CRaC](https://github.com/openjdk/crac/tree/crac) or future versions of OpenJDK. Package `org.crac` is a facade that either forwards the invocation to actual implementation or provides a no-op implementation. 

The `beforeCheckpoint(...)` and `afterRestore(...)` notifications are intended to be run by arbitrary thread, I hope there's no limitation in `close()` and `init()`.


jesperpedersen:
What is the JVM version target for this ?

rvansa:
Even with the change in JDK the dependency would be needed in order to keep supporting older JVMs.

Would be a solution that would shade the package directly into the JAR more acceptable?

jesperpedersen:
This is a productized artifact, so maybe instrument instead ?

rvansa:
Does shading somehow block productization? The facade uses BSD2 license.

By instrumentation, do you mean keeping an external repository and bytecode-level patching Agroal? Technically this is an option, though the least favorable.

jesperpedersen:
No, but it is an extra dependency - and this is JDK 22+ material.

barreiro:
Suggestion: create new `AgroalDataSource.FlushMode.INIT` mode and leave CRaC to whatever holds the `AgroalDatasource`.

rvansa:
@barreiro Thanks for the suggestion, but IIUC while it would be possibly to hook from the consuming project (e.g. Agroal integration into Quarkus) the idea is to push any CRaC support as low as possible to provide the biggest fan-out to consuming projects while having the modifications in as few places as possible.

barreiro:
closing this for now as _wontfix_. we may get back to it later.
anyway, thanks for your contribution.

PR #67, opened 2023-05-17
Restore autoconfiguration with SB3

Croway:
Autoconfiguration does not work OOB in SB3 since `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` has to be used instead of `spring.factories`


graben:
Could be also committed to 1.x branch since available since sb 2.7.0

Croway:
> Could be also committed to 1.x branch since available since sb 2.7.0

No problem, I can cherry pick the commit, but the new structure is mandatory only from SB3+, in 2.7.x spring.factories are deprecated.

graben:
@jesperpedersen : IMHO not an enhancement, it's a bug ;)

PR #70, opened 2023-06-02
AG-209 - check a transaction is active only by status ACTIVE or MARK…

zhfeng:
…ED_ROLLBACK

### DIFF
--- a/agroal-narayana/src/main/java/io/agroal/narayana/NarayanaTransactionIntegration.java
+++ b/agroal-narayana/src/main/java/io/agroal/narayana/NarayanaTransactionIntegration.java
@@ -5,6 +5,7 @@ package io.agroal.narayana;
 
 import io.agroal.api.transaction.TransactionAware;
 import io.agroal.api.transaction.TransactionIntegration;
+import jakarta.transaction.SystemException;
 import org.jboss.tm.TxUtils;
 import org.jboss.tm.XAResourceRecovery;
 import org.jboss.tm.XAResourceRecoveryRegistry;
@@ -20,7 +21,10 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import static jakarta.transaction.Status.STATUS_ACTIVE;
 import static jakarta.transaction.Status.STATUS_COMMITTED;
+import static jakarta.transaction.Status.STATUS_COMMITTING;
+import static jakarta.transaction.Status.STATUS_MARKED_ROLLBACK;
 import static jakarta.transaction.Status.STATUS_NO_TRANSACTION;
 import static jakarta.transaction.Status.STATUS_ROLLEDBACK;
 import static jakarta.transaction.Status.STATUS_UNKNOWN;
@@ -68,7 +72,7 @@ public class NarayanaTransactionIntegration implements TransactionIntegration {
 
     @Override
     public TransactionAware getTransactionAware() throws SQLException {
-        if ( transactionRunning() ) {
+        if ( transactionActive() ) {
             return (TransactionAware) transactionSynchronizationRegistry.getResource( key );
         }
         return null;
@@ -77,7 +81,7 @@ public class NarayanaTransactionIntegration implements TransactionIntegration {
     @Override
     public void associate(TransactionAware transactionAware, XAResource xaResource) throws SQLException {
         try {
-            if ( transactionRunning() ) {
+            if ( transactionActive() ) {
                 if ( transactionSynchronizationRegistry.getResource( key ) == null ) {
                     transactionSynchronizationRegistry.registerInterposedSynchronization( new InterposedSynchronization( transactionAware ) );
                     transactionSynchronizationRegistry.putResource( key, transactionAware );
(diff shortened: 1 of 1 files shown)


agroal-narayana/src/main/java/io/agroal/narayana/NarayanaTransactionIntegration.java
```diff
@@ -103,12 +107,26 @@ public void associate(TransactionAware transactionAware, XAResource xaResource)
...
     @Override
     public boolean disassociate(TransactionAware connection) throws SQLException {
-        if ( transactionRunning() ) {
+        if ( transactionActive() ) {
             transactionSynchronizationRegistry.putResource( key, null );
         }
         return true;
     }
 
+    private boolean transactionActive() throws SQLException {
```
mmusgrov, line 116: So you have added this but kept the method `transactionRunning()`: what is the relationship between the two methods?

zhfeng: TBH, I have no idea. `transactionRunning` is only used here
https://github.com/agroal/agroal/blob/697da5a3d56c4a08a8e8e77217fbb0e99962b76e/agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java#L393-L400

mmusgrov: Also your new method is replacing the method `transactionRunning` which contains:

```
            if ( transaction == null ) {
                // AG-183 - Report running transaction when reaper thread attempts rollback
                return TxUtils.isTransactionManagerTimeoutThread();
            }
```
but in your replacement you just return false if ( transaction == null ). Why don't you call the same method when it is null?

zhfeng: > So if you change that reference then the method verifyEnlistment will invoke your new method.

> Was it intentional?

Yeah, `verifyEnlistment` is used in `transactionCommit` and `transactionRollback` which I think they are related to `LocalXAResource`. As far as I read the codes, it looks like a `XAResourceWrapper`  to let a local tranaction to participant in a 2PC transaction.

If we reference the `transactionActiveCheck` to the new `transactionActive` method, `verifyEnlistment` will throw an Exception with  `Enlisted connection used without active transaction` when doing `transactionCommit` or `transactionRollback` because narayana thinks  that a  transaction in `COMMITTING` phase is **NOT** active.

zhfeng: I think we need some guys from agroal to clarify what is the `transactionRunning` intention?

mmusgrov: @barreiro Please can you take a look at this PR?

barreiro: @zhfeng @mmusgrov the intention of `transactionRunning` is to prevent the connection to cross a transaction boundary (it can either be acquired within the scope or outside). For that purpose, the transaction is considered to be running during the commit/rollback phases to prevent commit of work outside the tx.

Your definition of `transctionActive` is different, and essentially means that the tx is 'usable' for client code.

My question is how do both concepts co-exist. For instance, does acquiring a connection during COMMITING phase means that it's life-cycle _should_ or _shouldn't_ be tied to the running transaction ??

zhfeng: Thanks @barreiro !

It is a good question. So @mmusgrov in this case, from narayana side the following methods are allowed in `transactionSynchronizationRegistry` during `COMMITING` phase? is there anything in JTA spec?

- `getResource( key )` 
- `putResource( key )`
- `registerInterposedSynchronization(Synchronization synchronization)`

I think when acpuiring a connection, it needs to get a `TransactionAware` from the current transaction?

mmusgrov: For `getResource` and `putResource` there has to be transaction associated with the calling thread and it needs to be active (`Status.STATUS_ACTIVE`) and not marked rollback only (`Status.STATUS_MARKED_ROLLBACK`). During the commit phase the spec says that afterCompletion callbacks will be invoked in an undefined context.

Interposed synchronizations cannot be added after the transaction has prepared.

barreiro:
hi @mmusgrov @zhfeng. I have commited 602487330510b2893308b51e2882852f138b6e8f with a test case for this issue using `JDBCStore`. a few conclusions:

1) connections obtained by `JDBCStrore` can't be enlisted with the transaction. there is the reason already mentioned in this PR that the `TransactionSynchronizationRegistry` is not available in certain transaction states. even if it were, since `JDBCStore` does explicit calls to `commit()`, this would not be allowed on an enlisted connection where commit/rollback is managed by the transaction.

2) the only option left is for connections to not be enlisted. this means that a different connection from the pool will be handled to `JDBCSTore`. there can be issues with the availability of this connection (this means possible deadlocks / timeouts). I don't think it's a major issue for this particular use case, but I still would like to mention it here.

3) we have to consider changes to the transaction callback for this connections. the current situation is that connections that are not enlisted can't be used in the context of a running transaction. I propose tho change this behavior that so that connections obtained when the transaction is running but not active can still be used as long as the transaction state does not change. 

I will provide a follow-up PR, based on this one, so that you can review my fix. thanks!

barreiro:
superseded by #72

PR #71, opened 2023-06-12
fix: agroalPoolDataSourceMetadataProvider could not work in some scenarios

oburgosm:
According to `DataSourceUnwrapper.unwrap` [javadoc](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/jdbc/DataSourceUnwrapper.html#unwrap(javax.sql.DataSource,java.lang.Class)), **target** type  should be an interface.

In some scenarios this unwrap could return _null_, so is better to use `io.agroal.api.AgroalDataSource` instead of `io.agroal.springframework.boot.AgroalDataSource`

### DIFF
--- a/agroal-api/pom.xml
+++ b/agroal-api/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.agroal</groupId>
         <artifactId>agroal-parent</artifactId>
-        <version>2.2</version>
+        <version>2.3-SNAPSHOT</version>
     </parent>
     <artifactId>agroal-api</artifactId>
     <name>Agroal API</name>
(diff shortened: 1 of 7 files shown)


jesperpedersen:
@oburgosm Please, rebase and squash the commits

jesperpedersen:
@oburgosm Now, you have 5 commits

jesperpedersen:
@barreiro Sign-off, please

graben:
What are the preferred import order? I see them messed up!

barreiro:
Merged upstream as AG-211. Thanks for your contribution @oburgosm

PR #74, opened 2023-07-23
AG-213: Add micrometer implementation to promote metrics

graben:


barreiro:
awesome contribution !! thanks a lot @graben

PR #79, opened 2023-07-31
AG-216: Fine tune deploy step to run only on main repo

graben:
@barreiro : Small fine-tuning for AG-216


barreiro:
@graben thanks!

PR #80, opened 2023-08-16
AG-217: Add JNDI binding for connectable DataSource instances

graben:
Initial implementation that work well with [simple-jndi](https://github.com/h-thurow/Simple-JNDI) and Jetty JNDI via explicit `AgroalDataSourceJndiBinder` implementation.

AgroalDataSourceJndiBinder:
```
@Configuration
@AutoConfigureBefore(AgroalDataSourceConfiguration.class)
public class JettyAgroalDataSourceJndiBinderConfiguration {

    @Bean
    public AgroalDataSourceJndiBinder agroalDataSourceJndiBinder() {
        return (jndiName, dataSource) -> {
            try {
                new org.eclipse.jetty.plus.jndi.Resource(null, jndiName, dataSource);
                return true;
            } catch (NamingException ex) {
                return false;
            }
        };
    }
}
```

**Actually not working with Tomcat JNDI implementation!**

### DIFF
--- a/agroal-spring-boot-starter/pom.xml
+++ b/agroal-spring-boot-starter/pom.xml
@@ -36,10 +36,18 @@
             <artifactId>spring-tx</artifactId>
             <scope>provided</scope>
         </dependency>
+        <!-- Micrometer -->
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-core</artifactId>
             <version>${version.io.micrometer}</version>
+            <scope>provided</scope>
+        </dependency>
+        <!-- SLF4J -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${version.org.slf4j}</version>
             <scope>provided</scope>
         </dependency>
         <!-- JUnit (somehow surefire plugin gets pulled in) -->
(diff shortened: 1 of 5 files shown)


graben:
@barreiro : Could you plz have a look on this? Maybe you'll know a solution for Tomcat JNDI. ;-)

jesperpedersen:
We should look at making `java.util.logging.Logger` the default

graben:
Better slf4j? This needs a bit more of changing SB starter module.

jesperpedersen:
Logging should be unified - `git grep Log` to get you started

graben:
@jesperpedersen : Using JDK logger is a big step backwards since jdk logger has less feature than even commons logging!

graben:
Actually JDK logger is only used in unit tests (excluding DataSource implementing methods!) I would stull favor switching completly to slf4j as facade. Better for using in SpringBoot, Quarkus and WildFly!

graben:
Just to show how logging looks better with slf4j a quick migration for the starter. Yet to change the test module from jul to slf4j.

graben:
That's why I suggest to switch from commons-logging to slf4j. Included in last commit of this PR. JUL is still used in test module and could be changed also or left as is.

jesperpedersen:
Everything should be the same

graben:
Okay, will switch test module to slf4j, too. Will take some time, it's a lot of files.

barreiro:
there is no need to have the same logging in different modules IMO. 

these have different purposes, the logging in spring is when running in the context of that framework, while for tests is a quick way to check what's going on when there is a failure. 

I say leave it like it is WRT logging.

graben:
@barreiro : I think this PR is ready to merge

barreiro (approved):

barreiro:
@graben sorry for the delay and thanks for the contribution!

PR #81, opened 2023-09-14
AG-219 - Optimize resource tracking

barreiro:
Get rid of collections that are currently used to track JDBC resources (`Connection`, `Statement`, `ResultSet`) and instead have the wrapper objects that Agroal uses to track the resources form a collection between themselves.

This is implemented by having the wrapper extend the `AutoCloseableElement` class that implements the list (actually, it has stack behavior - FILO). There is also a special extension of the class used as head of the list. Once a new wrapper is created it tries to insert itself between the head and the first element of the list (if it exists). There is the invariant that at any given point in time the list can be traversed from the head and all wrappers are reachable.

Unlike before, the wrappers do not remove themselves during operation. They are dropped just at the moments when the list is cleared. This is not a big deal as the wrapped JDBC resources are still dropped on calls to the `close()` method.


jesperpedersen (approved):

franz1981 (changes requested):

franz1981:
LGTM eh, but a few lines of desc would help IMO

PR #83, opened 2023-09-15
AG-220 - Do not register connectable datasource for XA Recovery

barreiro:


graben:
Should have caught this earlier. ;-)

jesperpedersen (approved):

PR #76, opened 2023-07-27
AG-215: Fix SpringBoot related configuration bug and add spring to surefire groups

graben:
Ready once narayana-spring-boot 3.0.0 is released. (snowdrop/narayana-spring-boot#105) Actually, a SNAPSHOT version of narayana-spring-boot-starter is needed to run. Therefor, actions are currently failing!

PR #86, opened 2023-10-22
AG-223: autoconfigurations and orders

ronaldmik:
The autoconfigurations have been changed to ue the @AutoConfiguration annotation and also use explicit befores and afters. This ensures that agroal-spring-boot-starter's autoconfigurations are being triggered in the correct order in relation to spring boot's autoconfigurations.

Specifically, AgroalDataSourceConfiguration is going before DataSourceAutoConfiguration, so that spring boot does not create a DataSource because a DataSource bean is missing when AgroalDataSourceConfiguration also creates a DataSource when a DataSource bean is missing, making it a race condition. AgroalDataSourcePoolMetricsAutoConfiguration in turn will go after AgroalDataSourceConfiguration, because it is conditional on an AgroalDataSource bean being registered. AgoralDataSourcePoolMetricsAutoConfiguration will also trigger after CompositeMeterRegistryAutoConfiguration which seems to be the last autoconfiguration in spring boot actuator to register a MeterRegistry bean. While more libraries may actually create MeterRegistry beans, spring boot actuator is a known provider of MeterRegistry beans. Other libraries may very well register their MeterRegistry beans explicitly before spring boot actuator's CompositeMeterRegistryAutoConfiguration triggers. This seems to make CompositeMeterRegistryAutoConfiguration a good known candidate to autoconfige 'after'.

It seems like AgroalPoolDataSourceMetadataProviderConfiguration should remain a 'normal' @Configuration that needs to be imported by AgroalDataSourceConfiguration, just like in spring boot DataSourcePoolMetadataProvidersConfiguration is being imported by DataSourceAutoConfiguration.

### DIFF
--- a/agroal-spring-boot-starter/pom.xml
+++ b/agroal-spring-boot-starter/pom.xml
@@ -36,12 +36,18 @@
             <artifactId>spring-tx</artifactId>
             <scope>provided</scope>
         </dependency>
+        <!-- Actuator -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-actuator-autoconfigure</artifactId>
+            <optional>true</optional>
+        </dependency>
         <!-- Micrometer -->
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-core</artifactId>
             <version>${version.io.micrometer}</version>
-            <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <!-- SLF4J -->
         <dependency>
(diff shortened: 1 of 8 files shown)


barreiro:
@graben can I get your opinion on this, since you've contributed a lot in this area. thanks!

graben:
@barreiro sure, I'll have a look at it as soon as possible.

graben:
@barreiro : Overall, the proposed PR looks good. Did a few tests with my local use cases to test functionality and also tested with coming update #76

barreiro (approved):

barreiro:
merged. thank you both!

graben:
Well, the changes I proposed are not done. I'll fix it with my open branch.

PR #89, opened 2023-11-15
AG-225 Fix wrong exception message

graben:
@barreiro : Fix for the wrong exception message I suggested on initial PR


barreiro:
thanks for spotting!

PR #90, opened 2023-11-19
AG-226: Added a fix/'hack' in order to set the timer to a resolution of 1 ms for Windows, so that the BasicConcurrencyTests can succeed on Windows.

ronaldmik:
AG-226: Added a fix/'hack' in order to set the timer to a resolution of 1 ms for Windows, so that the BasicConcurrencyTests can succeed on Windows.

### DIFF
--- a/agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
+++ b/agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
@@ -53,6 +53,7 @@ public class BasicConcurrencyTests {
     @BeforeAll
     static void setup() {
         registerMockDriver();
+        Utils.windowsTimerHack();
     }
 
     @AfterAll
--- a/agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
+++ b/agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
@@ -69,7 +69,7 @@ public class BasicConcurrencyTests {
     @DisplayName( "Multiple threads" )
     @SuppressWarnings( "ObjectAllocationInLoop" )
     void basicConnectionAcquireTest() throws SQLException {
-        int MAX_POOL_SIZE = 10, THREAD_POOL_SIZE = 32, CALLS = 50000, SLEEP_TIME = 1;
+        int MAX_POOL_SIZE = 10, THREAD_POOL_SIZE = 32, CALLS = 50000, SLEEP_TIME = 1, OVERHEAD = 1;
 
         ExecutorService executor = newFixedThreadPool( THREAD_POOL_SIZE );
         CountDownLatch latch = new CountDownLatch( CALLS );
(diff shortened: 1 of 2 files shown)


agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
```diff
@@ -53,6 +53,7 @@ public class BasicConcurrencyTests {
     @BeforeAll
     static void setup() {
         registerMockDriver();
+        Utils.windowsTimerHack();
```
graben, line 56: Maybe only execute on windows? (`System.getProperty("os.name").startsWith("Windows")`)

agroal-test/src/test/java/io/agroal/test/basic/Utils.java
```diff
@@ -0,0 +1,22 @@
+package io.agroal.test.basic;
```
graben, line 1: Put header
```
// Copyright (C) 2023 Red Hat, Inc. and individual contributors as indicated by the @author tags.
// You may not use this file except in compliance with the Apache License, Version 2.0.
```

graben:
@ronaldmik : Nice catch, I disabled the test case locally because of that :-D

agroal-test/src/test/java/io/agroal/test/basic/BasicConcurrencyTests.java
ronaldmik: I added a Utils.isWindowsOS() method to check if the OS is Windows. Perhaps later this should be done a bit nicer if and when the necessity arises.

agroal-test/src/test/java/io/agroal/test/basic/Utils.java
ronaldmik: I added this header in the PR

ronaldmik:
@graben Yeah, I had already contacted Luis about this. But it did not stop annoying me! So I started digging into it and was quite surprised what I found :)

graben:
Well, even with this patch, it fails on my Windows machine. Seems a bit trickier than expected.

ronaldmik:
Hey @graben, on my machine it structurally succeeds since this fix. These kinds of tests may however always remain a bit tricky since they are not really deterministic.
Could you please test how much time the test takes in total when you are not using a timeout and then rerun the test with the patch in order to verify that it does run (a lot) faster with the patch? I expect the test to be a lot faster, but maybe it just isn't always fast enough and we should increase the margin of 1.5 to something a bit more.

graben:
It's really much faster, just added another 1000 to waitTime and the test succeeds. Maybe it is just this bit more needed.

ronaldmik:
Good to hear @graben! So the fix does work, but 1.5 times SLEEP_TIME * CALLS is not (always) enough margin. In a perfect hypothetical scenario without any overhead the test would take SLEEP_TIME * CALLS, but this is of course impossible. I feel like the formula for the waitTime should be something more like (SLEEP_TIME + OVERHEAD) * CALLS, where OVERHEAD is an arbitrary amount of time that serves as a margin for the overhead of switching, parking threads and other code that is being executed besides the 'sleep'. With OVERHEAD set to 1ms the test should already easily be succeeding and I don't feel like it is taking away the meaningfulness of the test. Perhaps @barreiro has an opinion about this?

barreiro:
very interesting! @ronaldmik brilliant investigation!

I'm fine with a change in the timeout for the test, if that increases it's reliability. the main point of the test is to ensure that the pool works under contention and no call hags. the timeout is necessary, but it can be made longer.

so far I have not seen any problems with the test, but I don't have a windows machine to test. I'll look into adding a windows test environment to CI, in a different issue.

PR #94, opened 2023-12-30
AG-228 - Not all (XA)DataSource's require an url to be set

graben:


ronaldmik:
@graben looks good to me

PR #95, opened 2024-01-08
AG-227 - Make RecoveryXAResource reconnectable

graben:


barreiro:
@graben this is great! 

we need to have a test case as well. I have yet been able to come up with a scenario where the `RecoveryManager` triggers the issue.

graben:
@barreiro : Do you have something to share?

graben:
@barreiro: Maybe I'll find a solution, let me think about it...

barreiro:
@graben I have something using a custom `RecoveryModule` for testing purposes. I'm unsure if that mimics the actual issue, but at least it can make possible to test what we think that may be happening.

graben:
@barreiro : Well, I asked your colleague for more details. Hope it would help to define a test case.

graben:
@barreiro : Lets's go with your test case.

PR #97, opened 2024-01-31
AG-233 - Add connection created and connection destroyed to pool interceptor

barreiro:


agroal-api/src/main/java/io/agroal/api/AgroalPoolInterceptor.java
```diff
@@ -29,6 +29,13 @@ public int compare(AgroalPoolInterceptor i1, AgroalPoolInterceptor i2) {
         }
     };
 
+
+    /**
+     * This callback is invoked when a new connection is created.
+     */
+    default void onConnectionCreated(Connection connection) {
```
graben, line 36: Better call onConnectionCreate, same tempus as return and destroy

agroal-pool/src/main/java/io/agroal/pool/util/InterceptorHelper.java
```diff
@@ -19,6 +19,16 @@ public final class InterceptorHelper {
     private InterceptorHelper() {
     }
 
+    public static void fireOnConnectionCreatedInterceptor(List<? extends AgroalPoolInterceptor> interceptors, ConnectionHandler handler) throws SQLException {
```
graben, line 22: Create vs Created

PR #100, opened 2024-02-02
AG-227: Explicitly close connection after reconnect for commit

graben:
Hi @barreiro, I think we should add this additional commit to avoid a new introduced leak. Maybe add the same pattern to other methods too. IMHO, none of them is used outside recover START/ END cycle. (Official Narayana modules, of course!)

### DIFF
--- a/agroal-narayana/src/main/java/io/agroal/narayana/RecoveryXAResource.java
+++ b/agroal-narayana/src/main/java/io/agroal/narayana/RecoveryXAResource.java
@@ -22,77 +22,91 @@ public class RecoveryXAResource implements AutoCloseable, XAResourceWrapper {
     public RecoveryXAResource(ResourceRecoveryFactory factory, String name) throws SQLException {
         connectionFactory = factory;
         jndiName = name;
-        connect();
+        connect( false );
     }
 
-    private void connect() throws SQLException {
+    private boolean connect(boolean flag) throws SQLException {
         if ( wrappedResource == null ) {
             xaConnection = connectionFactory.getRecoveryConnection();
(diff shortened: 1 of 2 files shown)


barreiro:
@graben right. I was thinking the same when I merged the PR, but was not able to spend any time with that.

agroal-narayana/src/main/java/io/agroal/narayana/RecoveryXAResource.java
```diff
@@ -53,6 +55,11 @@ public Xid[] recover(int flag) throws XAException {
     @Override
     public void commit(Xid xid, boolean onePhase) throws XAException {
         getConnectedResource().commit( xid, onePhase );
+        // Commit can be called after a recover(TMENDRSCAN) has been invoked.
+        // Close after reconnect to avoid a connection leak.
+        if ( reconnected ) {
```
barreiro, line 60: need the same for `rollback()`, no !?

graben: I did not find a reference on the first check, but double check is on my to-do list. Probably also add another test case which does the real whole roundtrip, but with more Narayana insights, of course.

graben:
@barreiro : I add another test class with an enhanced recovery test case maybe merge into RecoveryTests but it's quite unreadable with so many embedded classes.

graben:
@barreiro : So, I think this could be a leak free implementation ;)

barreiro (approved):

barreiro:
Great work !! Thanks !!

PR #101, opened 2024-02-02
AG-234: Add getUrl() to io.agroal.springframework.boot.AgroalDataSource

ronaldmik:
AG-234: Add getUrl() to io.agroal.springframework.boot.AgroalDataSource to expose the url used in order for spring boot's LiquibaseAutoConfiguration to be able to derive a DataSource from an existing AgroalDataSource.


barreiro:
Thank you very much for your contribution !!

PR #102, opened 2024-04-09
AG-237: Ability to change DataSource settings after starting

liubagha:
Hi, Agroal Community!

Quartus supports database management via Agroal Connection Pool (https://quarkus.io/guides/datasource).

When initializing an application, Quarkus allows you to create unremovable beans, for example DataSource. https://quarkus.io/guides/cdi-reference#remove_unused_beans

The existing library functionality does not allow you to restart the connection with new credentials in an existing bean. This functionality would be useful when switching to a backup database in case of an accident.

Can we add this functionality to the new version?

 Example:

        Properties properties = new Properties();
        properties.setProperty(ReloadDataSourceUtil.AGROAL_JDBC_URL, Configuration.getConfig().getConfigValue("kc.db-url").getValue());
        properties.setProperty(ReloadDataSourceUtil.AGROAL_USERNAME, Configuration.getConfig().getConfigValue("kc.db-username").getValue());
        properties.setProperty(ReloadDataSourceUtil.AGROAL_PASSWORD, Configuration.getConfig().getConfigValue("kc.db-password").getValue());

        io.agroal.pool.DataSource dataSource = (io.agroal.pool.DataSource) Arc.container().instance(AgroalDataSource.class).get();
        dataSource.reloadDataSourceWithNewCredentials(properties, AgroalDataSource.FlushMode.ALL);

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/DataSource.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/DataSource.java
@@ -7,14 +7,25 @@ import io.agroal.api.AgroalDataSource;
 import io.agroal.api.AgroalDataSourceListener;
 import io.agroal.api.AgroalDataSourceMetrics;
 import io.agroal.api.AgroalPoolInterceptor;
+import io.agroal.api.configuration.AgroalConnectionFactoryConfiguration;
 import io.agroal.api.configuration.AgroalDataSourceConfiguration;
+import io.agroal.api.configuration.AgroalDataSourceConfiguration.DataSourceImplementation;
+import io.agroal.api.configuration.supplier.AgroalConnectionFactoryConfigurationSupplier;
+import io.agroal.api.configuration.supplier.AgroalConnectionPoolConfigurationSupplier;
+import io.agroal.api.configuration.supplier.AgroalDataSourceConfigurationSupplier;
+import io.agroal.api.security.NamePrincipal;
+import io.agroal.api.security.SimplePassword;
+import io.agroal.pool.util.ReloadDataSourceUtil;
 
 import java.io.PrintWriter;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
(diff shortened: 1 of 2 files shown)


barreiro:
hi @liubagha. 

if I understand correctly, you want the pool be recreated with a new URL that uses the backup instance. I'm not sure if it's the best way to handle fail-over, from an architectural standpoint.  

I personally feel a bit uncomfortable with changing the URL of a running pool, and I believe the fail-over should be handled externally. 

@jesperpedersen I would also like to hear your opinion on this.

liubagha:
@barreiro, I would like to highlight the additional use case. 

There are additional circumstances beyond the [backup and recovery best practice](https://www.ibm.com/docs/en/configurepricequote/9.4.0?topic=practices-backup-recovery-best). May be it will be important.
If [database credentials have been compromised, or following password policies, we must change password(s) every N days](https://www.auditboard.com/blog/nist-password-guidelines/). Considering the features of the Quarkus platform, we don't have an alternative to reloading the application. This may have an impact on how the system will be maintained in the prom.

yrodiere:
IMO if you need to change DB credentials without downtime, you should use have a load-balancer in front of your app, allowing you to simply spawn a new instance of the application, and to redirect the load-balancer to that new instance once it's up.

Renewing the connection pool is something I could be convinced makes sense in some edge cases where restarting the app is not an option, but I'd be surprised if this wasn't already possible.

What I completely disagree with, on the other hand, is an approach allowing anyone to programmatically pass secrets like you're suggesting in this PR.
The right way to do this IMO would be through credential providers. In your app configuration, you wouldn't specify credentials directly, but through a credentials provider (e.g. using [Vault](https://docs.quarkiverse.io/quarkus-vault/dev/vault-datasource.html)). Then when you need an update:

1. Update the DB to allow new credentials.
2. Update the credentials in the source of information used by your credential provider (e.g. Vault).
3. Somehow get the application to renew its connection pool, thereby retrieving the new credentials and opening new connections using these.
4. Update the DB to disallow the old credentials.

barreiro:
closing this PR as there is an agreement that this not the best way to achieve to handle the fail-over. 

anyway, thanks for your contribution and for starting the discussion.

PR #103, opened 2024-04-30
AG-238 add connection test

marcosgopen:
https://issues.redhat.com/browse/AG-238
Add a test with two datasources  and Narayana XARecoveryModule to check all connections are closed.
This PR will be on Hold until https://github.com/jbosstm/narayana/pull/2246#pullrequestreview-2028657623 is merged and integrated into Agroal.
FYI @barreiro


mmusgrov:
FYI I ran Marco's test with and without the Narayana changes he proposed in naryana PR (#2246) and it passes with #2246 and fails without it. Therefore this is a good regression test.

graben:
@barreiro: I think this one is obsolete since 2.5 and could be closed.

PR #104, opened 2024-04-30
Delay opening of recovery connection

barreiro:
@mmusgrov this addresses some of the points discussed in the issue

### DIFF
--- a/agroal-test/src/test/java/io/agroal/test/narayana/RecoveryTests.java
+++ b/agroal-test/src/test/java/io/agroal/test/narayana/RecoveryTests.java
@@ -17,6 +17,8 @@ import io.agroal.narayana.NarayanaTransactionIntegration;
 import io.agroal.test.MockXAConnection;
 import io.agroal.test.MockXADataSource;
 import io.agroal.test.MockXAResource;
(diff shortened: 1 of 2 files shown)


graben:
Hi together,

would be interesting to test those changes together with the new test from #103 and the fix from jbosstm/narayana#2246 changed to do `.recover(XAResource.TMSTARTRSCAN | XAResource.TMENDRSCAN)` combined

mmusgrov (approved):
Thanks, we will change the way we call recover (after checking that our other existing integrations have bitwise support for the flags). Tagging @marcosgopen since he recently made some changes to when we recover XAResources.

agroal-test/src/test/java/io/agroal/test/narayana/RecoveryTests.java
```diff
@@ -215,6 +220,38 @@ void xaResourceReopenTest() throws SQLException {
         }
     }
 
+    @Test
+    @DisplayName( "Eager recovery module test" )
+    void EagerRecoveryModuleTest() throws SQLException {
```
graben, line 225: eagerRecoveryModuleTest insteadof EagerRecoveryModuleTest ;-)

PR #105, opened 2024-05-14
AG-239 - Recovery connections from the pool

barreiro:


agroal-pool/src/main/java/io/agroal/pool/wrapper/XAConnectionWrapper.java
```diff
@@ -0,0 +1,129 @@
...
+            trackedXAResources.closeAllAutocloseableElements();
+            handler.transactionEnd();
+        }
+    }
+
+    @Override
+    public XAResource getXAResource() throws SQLException {
+        try {
+            handler.traceConnectionOperation( "getXAResource()" );
+            handler.verifyEnlistment();
```
graben, line 86: @barreiro : Does this mean recovery needs an enlisted connection?

barreiro: @graben the opposite. it ensures that the connection was not enlisted somehow.

graben: I see, the method name is a bit confusing. Does contain throwing for several internal states.

graben (approved):

agroal-pool/src/main/java/io/agroal/pool/wrapper/XAResourceWrapper.java
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+    private static final XAResource CLOSED_XA_RESOURCE = (XAResource) newProxyInstance( XAResource.class.getClassLoader(), new Class[]{XAResource.class}, CLOSED_HANDLER );
+
+    // --- //
+
+    private final ConnectionHandler handler;
+    private XAResource wrappedXAResource;
+
+    public XAResourceWrapper(ConnectionHandler connectionHandler, XAResource resource, boolean trackJdbcResources, AutoCloseableElement head) {
```
graben, line 41: Did you add trackJdbcResources for potential future usage?

agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java
```diff
@@ -94,6 +95,10 @@ public ConnectionHandler(XAConnection xa, Pool pool) throws SQLException {
         touch();
     }
 
+    public XAConnectionWrapper xaConnectionWrapper() {
+        return new XAConnectionWrapper( this, xaConnection, connectionPool.getConfiguration().connectionFactoryConfiguration().trackJdbcResources() );
```
graben, line 99: Why not add a getter for XAConnection to ConnectionHandler  like rawConnection and getXaResource and let wrapper class call it from handler like the other wrappers?

agroal-pool/src/main/java/io/agroal/pool/wrapper/XAResourceWrapper.java
barreiro: yes, and to pass it down to the `ConnectionWrapper` when `getConnection()` is invoked

agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java
barreiro: that was an option, but don't want to expose the XAConnection if there is no need to do so

agroal-pool/src/main/java/io/agroal/pool/wrapper/XAResourceWrapper.java
```diff
@@ -0,0 +1,121 @@
...
+import java.sql.SQLException;
+
+import static java.lang.reflect.Proxy.newProxyInstance;
+
+/**
+ * @author <a href="lbarreiro@redhat.com">Luis Barreiro</a>
+ */
+public class XAResourceWrapper extends AutoCloseableElement implements XAResource {
+
+    private static final String CLOSED_HANDLER_STRING = XAConnectionWrapper.class.getSimpleName() + ".CLOSED_XA_RESOURCE";
```
graben, line 19: should be XAResourceWrapper.class I think ;)

agroal-pool/src/main/java/io/agroal/pool/wrapper/XAResourceWrapper.java
graben: but there is no getConnection on [XAResource](https://docs.oracle.com/en/java/javase/21/docs/api/java.transaction.xa/javax/transaction/xa/XAResource.html)?

barreiro: you're right. I was confusing with XAConnection.

PR #107, opened 2024-05-22
AG-241 - Switch from property injection to constructor injection

graben:
Hi @barreiro, I'm actually working on a poc for a multiple dataSource SpringBoot application. It would be nice to have the property injection in AgroalDataSourceConfiguration to be switched to constructor injection. Make it a bit nicer to use. If okay, I would open a JIRA issue case.

Btw, better name would be AgroalDataSourceAutoConfiguration. Rename as well?


barreiro:
yes, please open a tracking JIRA. you can rename the class too.

barreiro:
looks good. thanks !!

PR #109, opened 2024-06-24
AG-244 - Fix order of conditions for flushing connections

graben:


aschoerk:
this will restrict the pool size to one if flushOnClose is set. Is that intentional?

graben:
I wouldn't think so. returnConnectionHandler is called while return a borrowed connection to the pool. Pool sizing works asynchronous.

aschoerk:
Hmmm, the original cause for me to search this code was that pools where flushOnClose was set, never grew greater than minSize and returned exceptions as soon as minSize+1 was necessary.
I that the cause for that was this part of the code.

graben:
Well, I did a quick test with my own spring-boot test app and everything works as expected. Connections gets destroyed after usage and new connections are created till min size reached. But there might be conditions when pool is under heavy load that refilling might be not fast enough!

Btw have to fill a gap in the sb starter ;)

barreiro:
looks good to me !!

PR #111, opened 2024-12-30
Spring Boot Starter: enable setting of security provider and credentials

grmblfrz:
We are using vault for db password rotation. for this to work without restarting the application (which would break long running jobs) we need to be able to supply a suitable security provider with corresponding credentials when creating the data source. This is not possible at the moment. We would like to do something like this:
```
AgroalDataSource ds = dataSourceProperties().initializeDataSourceBuilder()
                                            .type(AgroalDataSource.class)
                                            .build();
ds.addSecurityProvider(vaultSecurityProvider);
ds.credential(vaultCredentials);
```
This pull request adds those methods.


graben:
@grmblfrz: Can you please also provide test case(s) to proof and open a JIRA issue at https://issues.redhat.com to document.

@barreiro: What do you think, I actually struggle a bit how the feature is implemented. Supplier gets closed afterPropertiesSet is invoked. Should be changed to get more Spring-like style?

PR #117, opened 2025-02-20
AG-252 - Throw exception on acquire after tx timeout

barreiro:


agroal-narayana/src/main/java/io/agroal/narayana/NarayanaTransactionIntegration.java
```diff
@@ -82,17 +83,21 @@ public NarayanaTransactionIntegration(TransactionManager transactionManager, Tra
 
     @Override
     public TransactionAware getTransactionAware() throws SQLException {
-        if ( getTransactionPhase() == TRANSACTION_ACTIVE ) {
+        TransactionPhase phase = getTransactionPhase();
+        if ( phase == TRANSACTION_ACTIVE ) {
             return (TransactionAware) transactionSynchronizationRegistry.getResource( key );
+        } else if ( phase == TRANSACTION_ROLLED_BACK ) {
```
graben, line 89: What about STATUS_ROLLING_BACK?

barreiro: `STATUS_ROLLING_BACK` is a state where the transaction is considered active, so the TSR is available

graben (approved):

PR #118, opened 2025-03-18
AG-251 - Be cautious about exceptions in connection validators

barreiro:


graben (approved):

graben:
Maybe add logging to the catch throwable. Might help for further investigation.

graben:
@barreiro : go for it 😉

PR #120, opened 2025-04-23
AG-255 - Reset read-only state

barreiro:
This adds a new setting, `readOnly` that keeps connections in read-only state.


graben (changes requested):
I'm missing Spring Boot implementation of readonly flag.

PR #123, opened 2025-05-04
AG-257 - Add health check that use internal isHealthy method

graben:
@barreiro:  Please have a look at it.


barreiro:
LGTM. thanks !!

PR #124, opened 2025-05-15
Fix potential NPE in XAConnectionWrapper

beikov:


beikov:
Just found it myself:

* https://quarkus.io/guides/all-config#quarkus-agroal_quarkus-datasource-jdbc-detect-statement-leaks
* https://docs.wildfly.org/20/wildscribe/subsystem/datasources/data-source/ExampleDS/index.html#attr-track-statements

So it seems this NPE was not observed because by default, tracking is enabled.

It kind of feels a bit odd though, that Quarkus talks about just leak detection. The statement cancellation that you added with AG-231 is IMO something that makes the statement tracking a bit broader and useful beyond leak detection. Maybe the property should be modeled as an enum like in Wildfly.

barreiro:
fixed as [AG-260](https://issues.redhat.com/browse/AG-260)

PR #126, opened 2025-05-23
AgroalDataSourceHealthContributorAutoConfiguration supports wrapped DataSources

xtermi2:
Now the `AgroalDataSourceHealthContributorAutoConfiguration` can handle `AgroalDataSource` wrapped by another DataSource by using `DataSourceUnwrapper.unwrap`.

This is important e.g. when `org.springframework.jdbc.datasource.DelegatingDataSource` wraps `AgroalDataSource`. Currently `AgroalDataSourceHealthContributorAutoConfiguration` fails in this case with the assertion `Beans must not be empty`.


barreiro:
@xtermi2 thanks for reporting!

PR #127, opened 2025-05-30
more aggressively reducing state held after close

shawkins:
A possible implementation of https://github.com/quarkusio/quarkus/issues/48135

After reading the notes I see that self removal on close is not intended / desirable, so this is an attempt to further minimize the state held. The first commit eliminates the shadow variables for the Prepared / Callable statements and removes holding the AutoCloseableElement.newHead(). The second commit further combines these to just a single field.

Let me know if something like this is desirable and I'll create an issue for it.


barreiro:
@shawkins yes, for sure this is desirable. the only thing I'm divided is the introduction of the abstract class (for performance reasons). please create an issue for this.

shawkins:
Added another commit to use a concrete base class, and to hold all state in a single variable with the connection wrapper holding shared closed statements.

shawkins:
Opened #129 as a more targeted alternative. While the changes here help I realize they are quite large, and we do still accumulate memory entries. At the very least I'll still need to undo the auto organization of imports the IDE did and address any potential concurrency concerns. So if #129 seems more viable, then we can just close this for now.

PR #131, opened 2025-06-24
[Draft] contention removal

Sanne:
@barreiro would something like this be acceptable?

I'm seeing in our benchmarks a fair bit of noise related to these pool statistics, it's not much but I think it's worth doing something about it - especially when they are not being used.

N.B. sending as a draft as I haven't tested this yet: would prefer to hear your opinion before I spend too much time on it.


barreiro:
@Sanne I'm closing this PR as it's superseded by #134

PR #132, opened 2025-06-25
AG-262: Do not flush the connection on a unilateral rollback on rollback

ronaldmik:
AG-262: On rollback, when the RM return any of the XA_RB* codes, do not call TransactionAware#setFlushOnly


barreiro:
looks good to me !! thanks @ronaldmik

PR #133, opened 2025-07-31
AG-264 - Replace custom AgroalSynchronizer with LinkedTransferQueue

barreiro:


barreiro:
@franz1981 I would love to have your approval on this one.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -315,14 +326,22 @@ private ConnectionHandler handlerFromTransaction() throws SQLException {
...
         try {
-            for ( ; ; ) {
+            for ( int i = 0; ; i++ ) {
                 // If min-size increases, create a connection right away
                 if ( allConnections.size() < configuration.minSize() ) {
                     task = housekeepingExecutor.executeNow( new CreateConnectionTask() );
                 }
+                if ( i == 0 && handlerTransferQueue.hasWaitingConsumer() ) { // On the first iteration, block right away if there are other threads already blocked
+                    // There is a race condition here (if a thread was blocked but by the time poll is executed a transfer allready took place)
+                    // Because of that do not block for the whole remaining duration. Do it for at most a secong and then move on to perform a scan
```
graben, line 339: Typo "secong"

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -315,14 +326,22 @@ private ConnectionHandler handlerFromTransaction() throws SQLException {
...
+                    ConnectionHandler handler = handlerTransferQueue.poll( Long.min( SECONDS.toNanos( 1 ), remaining > 0 ? remaining * 9 / 10 : MAX_VALUE ), NANOSECONDS );
```
graben, line 340: "SECONDS.toNanos( 1 )" make it a constant to avoid computation repeatedly

graben (changes requested):
As far as I can evaluate, looks okay. Just two small comments.

graben (approved):

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -207,7 +219,7 @@ public void close() {
         allConnections.clear();
         activeCount.reset();
 
-        synchronizer.release( synchronizer.getQueueLength() );
+        while ( handlerTransferQueue.tryTransfer( TRANSFER_POISON ) ) ; // Unblock waiting threads with CancellationException
```
franz1981, line 222: this always expect some thread to be parked waiting for it; is it correct?
If there are no waiting thread this should just leave this, without looping?

franz1981 (changes requested):

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -319,14 +331,22 @@ private ConnectionHandler handlerFromTransaction() throws SQLException {
...
-            for ( ; ; ) {
+            for ( int i = 0; ; i++ ) {
                 // If min-size increases, create a connection right away
                 if ( allConnections.size() < configuration.minSize() ) {
                     task = housekeepingExecutor.executeNow( new CreateConnectionTask() );
                 }
+                if ( i == 0 && handlerTransferQueue.hasWaitingConsumer() ) { // On the first iteration, block right away if there are other threads already blocked
+                    // There is a race condition here (if a thread was blocked but by the time poll is executed a transfer allready took place)
+                    // Because of that do not block for the whole remaining duration. Do it for at most a second and then move on to perform a scan
+                    ConnectionHandler handler = handlerTransferQueue.poll( Long.min( ONE_SECOND, remaining > 0 ? remaining * 9 / 10 : MAX_VALUE ), NANOSECONDS );
```
franz1981, line 345: you can put this out of this loop (and not in the i == 0 within the loop) and executing it regardless without checking the number of waiting consumers?

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -403,7 +431,7 @@ private boolean performValidation(ConnectionHandler handler, ConnectionHandler.S
         fireBeforeConnectionValidation( listeners, handler );
         if ( handler.isValid() && handler.setState( VALIDATION, targetState ) ) {
             fireOnConnectionValid( listeners, handler );
-            synchronizer.releaseConditional();
+            handlerTransferQueue.tryTransfer( handler );
```
franz1981, line 434: maybe you want to offer it instead?
If a later consumer block after a "failed"  `tryTransfer` it won't find anything and will wait forever.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -488,7 +516,7 @@ public void returnConnectionHandler(ConnectionHandler handler) throws SQLExcepti
 
         if ( handler.setState( CHECKED_OUT, CHECKED_IN ) ) {
             // here the handler is already up for grabs
-            synchronizer.releaseConditional();
+            handlerTransferQueue.tryTransfer( handler );
```
franz1981, line 519: same here?

agroal-pool/src/main/java/io/agroal/pool/Poolless.java
```diff
@@ -426,7 +435,7 @@ private void flushHandler(ConnectionHandler handler) {
         handler.setState( FLUSH );
         allConnections.remove( handler );
         activeCount.decrementAndGet();
-        synchronizer.releaseConditional();
+        handlerTransferQueue.tryTransfer( handler );
```
franz1981, line 438: same here: why not offering to allow a late consumer to pick it?

franz1981 (changes requested):

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
barreiro: I see your point, and I've been thinking about it, but the thing is that the calls are not symmetric. an acquire is not required to call `poll()`  and if every connection return would call `offer()` there would be a huge leak.

`LinkedTransferQueue` is not bound either (that could be an option to limit the number of connections on the queue) and calling `size()` is an expensive operation.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
barreiro: it's not practical because of the code that comes before it on the loop

barreiro:
@MRLimcon thank you so much for running the tests and verifying there are no issues.

@franz1981 I have added a callback that is invoked before the thread blocks. with that I can verify that the same kind of race conditions exist before and with this PR, so I don't think there is a regression in that regard. we can try to improve further in next releases.

franz1981:
Agree @barreiro the point of this change was to fix the problem with virtual thread without breaking new stuff and it hits the spot..well done both ❤️

PR #134, opened 2025-07-31
AG-265 - Only collect pool statistics when needed

barreiro:


Sanne:
Excellent! thanks again :)

PR #142, opened 2025-12-05
AG-281 - Introduce `ConnectionValidator#sqlValidator`

gastaldi:
This introduces a new ConnectionValidator to allow specifying the validation SQL

### DIFF
--- a/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
+++ b/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
@@ -243,6 +243,25 @@ public interface AgroalConnectionPoolConfiguration {
             };
         }
 
+        /**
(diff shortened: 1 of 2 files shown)


agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -206,8 +207,8 @@ static ConnectionValidator defaultValidator() {
                 @Override
                 public boolean isValid(Connection connection) {
                     try {
-                        return connection.isValid( 0 );
```
barreiro: please respect the code style (space after `(` and before `)`). it's not enforced, but a good practice to keep.

gastaldi: That was formatted by the IDE, I'll change it

barreiro (approved):

PR #143, opened 2025-12-05
AG-280 - Add onConnectionCreationFailure callback

barreiro:


graben:
Do not forget the Spring boot starter. There is also an impl of the Listener.

graben:
Sorry I shouldn't read PRs on my mobile phone. 🫣

PR #144, opened 2025-12-05
AG-270: Add ability to get connection with readOnly flag been set immediately

graben:


graben:
@barreiro: I think this ready for review

PR #145, opened 2025-12-07
AG-281 - Add basic integration of sql connection validator into AgroalPropertiesReader

graben:
@gastaldi: Little extension to your last commit. Might need also to integrate with AgroalPropertiesReader.

@barreiro: Actually no idea how a good syntax can be looked

### DIFF
--- a/agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalPropertiesReader.java
+++ b/agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalPropertiesReader.java
@@ -264,7 +264,7 @@ public class AgroalPropertiesReader implements Supplier<AgroalDataSourceConfigur
      * <li>`empty` for the default {@link ConnectionValidator#emptyValidator()}</li>
      * <li>`default` for {@link ConnectionValidator#defaultValidator()}</li>
      * <li>`defaultX` for {@link ConnectionValidator#defaultValidatorWithTimeout(int)} where `X` is the timeout in seconds</li>
-     * <li>`custom[QUERY]X` for {@link ConnectionValidator#sqlValidator(String, int)} where `QUERY` is the sql query and `X` is the timeout in seconds</li>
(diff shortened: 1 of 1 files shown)


agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalPropertiesReader.java
```diff
@@ -267,6 +275,14 @@ public static ConnectionValidator parseConnectionValidator(String connectionVali
             return defaultValidator();
         } else if ( connectionValidatorName.regionMatches( true, 0, "default", 0, "default".length() ) ) {
             return defaultValidatorWithTimeout( (int) parseDurationS( connectionValidatorName.substring( "default".length() ) ).toSeconds() );
+        } else if ( connectionValidatorName.regionMatches( true, 0, "custom", 0, "custom".length() ) ) {
```
gastaldi: Why `custom` and not `sql`?

graben:
@gastaldi : Quite ugly to get proper test for an anonymous inner class. :-)

agroal-api/src/main/java/io/agroal/api/configuration/supplier/AgroalPropertiesReader.java
```diff
@@ -257,6 +264,7 @@ private void applyJdbcProperties(BiConsumer<? super String, ? super String> cons
      * <li>`empty` for the default {@link ConnectionValidator#emptyValidator()}</li>
      * <li>`default` for {@link ConnectionValidator#defaultValidator()}</li>
      * <li>`defaultX` for {@link ConnectionValidator#defaultValidatorWithTimeout(int)} where `X` is the timeout in seconds</li>
+     * <li>`custom[QUERY]X` for {@link ConnectionValidator#sqlValidator(String, int)} where `QUERY` is the sql query and `X` is the timeout in seconds</li>
```
gastaldi:
```suggestion
     * <li>`sql[QUERY]X` for {@link ConnectionValidator#sqlValidator(String, int)} where `QUERY` is the sql query and `X` is the timeout in seconds</li>
```

gastaldi:
We can add the tests in a separate PR if that takes too much time

PR #147, opened 2025-12-09
AG-274 - ConnectionCreateTimeout

### LINKED ISSUE #146: Database is available, but db pool is stuck when creating new connections
Just as a reference: https://issues.redhat.com/browse/AG-274

What is the main issue tracker for agroal? Or should I transfer the issue to github?

BTW: I am currently working on a PR

denniskniep:
Fixes: https://issues.redhat.com/browse/AG-274

**Issue TL;DR;**
Agroal pool can get “stuck” because all new connections are created by a single background thread, and a blocked DB connect call in that thread blocks the entire pool. 
In my observed scenario the blocking happens inside the MySQL driver during a socket read with no timeout, so if a connection attempt hangs, no further connections can be created.

**Solutions considered:**
Setting a socket timeout in the jdbc props. This helps, but has also an implication on the later usage of the connection. Because then also normal SQL Statements are limited to that timeout.

**Implementation Details:**
Therefore I decided to implement a general and database driver agnostic connection creation timeout.
It is implemented as an opt‑in feature: by default the timeout is set to 0, meaning no timeout and the behavior is identical to the current implementation. If you configure the connection creation timeout to a value greater than 0, the pool will attempt to create a connection and wait up to that timeout. If the timeout elapses without a successful connection, the creation task is canceled and a new attempt is started, repeating this process until the acquisition timeout is reached.

**Testing:**
For testing I decided against a Thread.sleep implementation to simulate a connection creation which takes longer. Because Thread.sleep can be interrupted easily. This is not true when blocked by reading from a socket. Therefore I implemented a small FakeServer using ServerSockets.

**Potential follow-up:**
Before I go ahead with further improvements I wanted to check with the maintainers, if that is the right direction to solve that issue. I think we can improve the feature by tracking metrics around how many connection creations failed or timed-out etc.

If you have any questions, let me know.

Fixes #146

### DIFF
--- a/agroal-api/src/main/java/io/agroal/api/AgroalDataSourceListener.java
+++ b/agroal-api/src/main/java/io/agroal/api/AgroalDataSourceListener.java
@@ -27,6 +27,11 @@ public interface AgroalDataSourceListener {
     default void onConnectionCreation(Connection connection) {}
 
     /**
+     * This callback is invoked when a new connection creation task times out and therefore canceled.
+     */
+    default void onConnectionCreationTimeout() {}
+
+    /**
      * This callback is invoked when a new connection can't be established.
      */
     default void onConnectionCreationFailure(SQLException sqlException) {}
--- a/agroal-test/pom.xml
+++ b/agroal-test/pom.xml
@@ -18,10 +18,16 @@
     </repositories>
     <dependencies>
         <dependency>
+            <groupId>com.mysql</groupId>
+            <artifactId>mysql-connector-j</artifactId>
+            <version>8.3.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.agroal</groupId>
             <artifactId>agroal-api</artifactId>
             <version>${project.version}</version>
-            <scope>test</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>io.agroal</groupId>
@@ -51,7 +57,7 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
             <version>${version.org.junit.jupiter}</version>
-            <scope>test</scope>
+            <scope>compile</scope>
         </dependency>
         <!-- OSGi -->
         <dependency>
--- a/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
+++ b/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
@@ -153,6 +153,17 @@ public interface AgroalConnectionPoolConfiguration {
      */
     void setAcquisitionTimeout(Duration timeout);
 
+    /**
+     * The maximum amount of time the agroal housekeeping thread may be blocked waiting for a new connection.
+     * A duration of {@link Duration#ZERO} means that a thread will wait indefinitely.
+     */
+    Duration connectionCreateTimeout();
+
+    /**
+     * Sets a new amount of time the agroal housekeeping thread may be blocked waiting for a new connection. Threads already blocked when this value changes do not see the new value when they unblock.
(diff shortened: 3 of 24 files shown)


agroal-test/pom.xml
```diff
@@ -227,6 +233,18 @@
...
+            <artifactId>agroal-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${version.org.junit.jupiter}</version>
+            <scope>compile</scope>
+        </dependency>
```
gastaldi: They are already declared above, maybe remove the test scope?

graben:
Overall, it seems there is one "big" bug. The executor is used for every housekeeping task. This change will introduce a timeout feature with logging not only for connection creation?

denniskniep:
> This change will introduce a timeout feature with logging not only for connection creation?

That´s a good point, but If I am not mistaken, than all `CreateConnectionTask` are started via `housekeepingExecutor.executeNow`. Which means they are added to the `priorityTasks` list. 
Normal housekeeping tasks are added via `housekeepingExecutor.schedule` (e.g. FlushTask, DestroyConnectionTask, etc.), except the shutdown.

Point is that before normal housekeeping tasks are executed, the `beforeExecute` is kicking in and run the PriorityTasks list.
But those priority tasks are only CreateConnectionTask. I added further code to reflect that fact also in the API.
(Shutdown is now implemented as internal special case) No one can anymore raise arbitrary priority tasks from outside. 


 > Overall, it seems there is one "big" bug. The executor is used for every housekeeping task.
 
I agree, but to solve that, it would imply a way larger code change and refactoring. Which would be of course fine, but we would need to discuss the approach first I guess.

graben:
> @graben thanks for your feedback.
> 
> > I would recommend such an approach only when using virtual threads.
> 
> Then we would need to lift the language level to at least Java19. Is that fine?
> 
> Do you think that a sub pool of virtual threads would be okay?
> 
> ```
> this.connectionCreationExecutor = Executors.newVirtualThreadPerTaskExecutor();
> ```
> 
> > Reuse the same pool with a priority runnable marker interface might help.
> 
> My understanding is that the current thread pool is limited to a single thread, means priorities are pointless, correct me if I am wrong with that assumption.
> 
> ```
> housekeepingExecutor = new PriorityScheduledExecutor( 1, "agroal-" + HOUSEKEEP_COUNT.incrementAndGet(), configuration.connectionCreateTimeout(), listeners );
> ```

Well, we switch from 11 to 17 with next major release 3.0. So I do not expect the next step forward to 21+ in the near future. We are following/aiming our target platforms Quarkus/Spring Boot.

Maybe we can think about doing a multi-jar with a virtual thread tech preview. Many projects did/are doing such approaches. But scheduling/threading stuff is a topic @barreiro is much deeper into.

denniskniep:
Hi @gastaldi,

> Sure, I was never against the solution proposed btw, just wanted to make sure we could reproduce it with less test code

Is `BlockingMockDriver` a solution for the tests that is fine for you? The behavior can now be replicated using minimal test code.

agroal-api/src/main/java/io/agroal/api/AgroalDataSourceListener.java
```diff
@@ -26,6 +26,11 @@ default void beforeConnectionCreation() {}
      */
     default void onConnectionCreation(Connection connection) {}
 
+    /**
+     * This callback is invoked when a new connection creation task is canceled.
+     */
+    default void onConnectionCreationCanceled() {}
```
gastaldi:
```suggestion
    default void onConnectionCreationTimeout() {}
```

barreiro:
I'm closing this one in favor of #149. I already highlighted some of the concerns, but the main reason is the general consensus here that the main responsibility for aborting the establishment of the connection is at the JDBC driver level.

PR #149, opened 2025-12-18
AG-274 - Retry connection establishment

barreiro:
This is in  alternative to #147 

[AgroalConnectionFactoryConfiguration#loginTimeout](https://github.com/agroal/agroal/blob/2.8/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionFactoryConfiguration.java#L39) already provides a timeout for connection establishment. having a similar timeout at the pool level makes little sense.

I went a step further and implemented a retry if there's still time available within the `acquisitionTimeout`.  I choose to retry only once (after a backoff period) but this could be made either configurable or extended to the whole duration of `acquisitionTimeout`.

I adapted tests from @denniskniep with cases where `acquisitionTimeout` is not set (no retry) and also when it's set.

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -351,6 +351,7 @@ public final class ConnectionPool implements Pool {
         long remaining = configuration.acquisitionTimeout().toNanos();
         long deadline = remaining > 0 ? nanoTime() + remaining : MAX_VALUE;
         Future<ConnectionHandler> task = null;
+        ExecutionException createException = null;
         try {
             for ( int i = 0; ; i++ ) {
                 // If min-size increases, create a connection right away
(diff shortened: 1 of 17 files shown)


graben (approved):

denniskniep:
Hi @barreiro ,
your proposal using `loginTimeout` in `WhenServerAcceptsAndDoNotRespond_ThenAcquisitionTimeoutAndConnectionCreationHangs` solves the issue indeed in the unit test.
```
.connectionPoolConfiguration( cp -> cp
  .maxSize( 1 )
  .acquisitionTimeout( Duration.ofSeconds( 1 ) )
  .connectionFactoryConfiguration( cf -> cf.loginTimeout( Duration.ofSeconds( 1 ) ))
 );
```
But the unit test was only an attempt to reproduce the issue via a test case - its not 100% the same as in my particular environment. I already tried to solve the issue in my environment by setting the `loginTimeout`, but that didn´t solved it.

I will try to write a test that reproduces it even closer.

PR #150, opened 2025-12-21
AG-274 - Troubleshooting connection creation

denniskniep:
Hi @barreiro, @graben, @gastaldi 

This is on top of @barreiro´s PR: [AG-274 - Retry connection establishment](https://github.com/agroal/agroal/commit/3452ece4e5901b7b667d260590d5fc169812c0e8)

I wrote a unit test which confirms, that even when `loginTimeout` is set, the connection creation is **blocked** by a single connection creation thread. When this situation happens, then Agroal pool is “stuck” because the blocked connection creation call in that thread blocks the entire pool. Please see: `WhenLoginTimeoutSetAndCommandExecutionDoesNotComplete_ThenAcquisitionTimeoutAndConnectionCreationHangs`

I improved the MySql Fake Server implementation. Now it handles the initial handshake between client and server and basic commands (If we want to keep this implementation in the codebase, then I will improve the code quality. The current state is just for showcasing the issue!)

Here is the explainer why  `loginTimeout` does not help in that particular case:

During connection creation, the `loginTimeout` **only** covers the phase from initial handshake to successful authentication.
JDBC driver executes further commands after successful authentication during the connection creation:
ie.: `SELECT @@sql_mode AS sql_mode`, `SET autocommit=1`, `SET sql_mode='STRICT_TRANS_TABLES' `, etc.

Those queries are not covered by the `loginTimeout`. That means, that every of those post-authentication queries during the connection creation can potentially lead to a stuck pool. The ServerBehaviour `CommandExecutionDoesNotComplete` simulates exactly that inside the mentioned unit test.

Additional note:
The unit test `WhenLoginTimeoutSetAndLoginDoesNotComplete_ThenAcquisitionTimeoutAndConnectionCreated` proves that the `loginTimeout` has an effect, but only if the connection creation is not proceeding in the phase between initial handshake and successful authentication. We can simulate this with the ServerBehaviour `LoginDoesNotComplete`

### DIFF
--- a/agroal-test/src/main/java/io/agroal/test/MySQLConnectMockDriver.java
+++ b/agroal-test/src/main/java/io/agroal/test/MySQLConnectMockDriver.java
@@ -12,6 +12,7 @@ import java.sql.Driver;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.Properties;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 public class MySQLConnectMockDriver implements MockDriver {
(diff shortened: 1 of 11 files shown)


barreiro:
I'm closing AG-274 for now, but I have some ideas for changes to address the single-threaded connection establishment.

I'll try to come up with something in the near future. The test cases in this PR will be handy. thanks for the contribution.

barreiro:
@denniskniep PR for the new feature is #163. I'd love to have your feedback too.

PR #151, opened 2025-12-23
AG-276 - Make DestroyConnectionTask as a priority if flushOnClose is configured

graben:
Hi @barreiro,

this is a quick idea to probably solve the AG-276. Actually, connections are created with priority, while destruction is only queued for later processing. IMHO, under "heavy" load this can lead to connections been created until database limit reached because the former used one still exists. Therefore, I propose to do a executeNow for destruction in these cases. The flow will change a bit as destruction has priority while fill task does not. But I think that's the right way to get the new connection created after the old one has been destroyed to keep processing under limit.

What do you think?


graben:
@barreiro : undrafted 😀

PR #173, opened 2026-03-11
AG-299: Add option to add interceptors to pool through pool configuration using spring boot

ronaldmik:


agroal-spring-boot-starter/src/main/java/io/agroal/springframework/boot/AgroalDataSourceAutoConfiguration.java
```diff
@@ -119,6 +123,8 @@ public AgroalDataSource dataSource(
         credentials.forEach( dataSource::addCredential );
         recoveryCredentials.forEach( dataSource::addRecoveryCredential );
 
+        poolInterceptor.stream().forEach( dataSource::addPoolInterceptor );
```
graben: I think stream has no benefit here, can call foreach directly

graben (approved):

barreiro (approved):

PR #179, opened 2026-03-25
AG-304 -  Fix interleaving issue with the Narayana reaper thread 

marcosgopen:
For MSSQL the reaper rollback does not rollback the SQL executed by the prepareStatement as its connection is in local mode. 

`mvn clean install -Pit -Dtest=MSSQL*` from the agroal-integration-tests folder

### DIFF
--- a/agroal-integration-tests/src/test/java/io/agroal/tests/xa/MSSQLXAReaperRaceTest.java
+++ b/agroal-integration-tests/src/test/java/io/agroal/tests/xa/MSSQLXAReaperRaceTest.java
@@ -1,0 +1,71 @@
+package io.agroal.tests.xa;
+
+import java.util.logging.Logger;
+
+import org.jboss.byteman.contrib.bmunit.BMUnitConfig;
+import org.junit.jupiter.api.Tag;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.mssqlserver.MSSQLServerContainer;
+
+/**
+ * XA reaper race integration test against MSSQL using Testcontainers.
+ * Uses WAITFOR DELAY for slow SQL that holds the read lock past the transaction timeout.
+ *
+ * MSSQL XA requires the JDBC XA stored procedures to be installed.
+ * The init script (mssql-xa-init.sql) installs them via sp_sqljdbc_xa_install.
+ */
+@Tag( "testcontainers" )
+@Testcontainers
+@BMUnitConfig(debug = true)
+class MSSQLXAReaperRaceTest extends XAReaperRaceTestBase {
+
+    private static final Logger logger = Logger.getLogger( MSSQLXAReaperRaceTest.class.getName() );
+
+    @Container
+    static MSSQLServerContainer mssql = new MSSQLServerContainer( System.getProperty( "mssql.testcontainer.image", "mcr.microsoft.com/mssql/server:2022-CU23-ubuntu-22.04" ) )
+            .acceptLicense()
+            .withInitScript( "mssql-xa-init.sql" );
+
+    @Override
+    String xaDataSourceClassName() {
+        return "com.microsoft.sqlserver.jdbc.SQLServerXADataSource";
+    }
+
+    @Override
+    JdbcDatabaseContainer<MSSQLServerContainer> container() {
+        return mssql;
+    }
+
+    @Override
+    String slowSQL() {
+        return "WAITFOR DELAY '00:00:04'";
+    }
+
+    /**
+     * MSSQL's MSDTC actively cancels the distributed transaction when the
+     * reaper fires, causing the in-flight WAITFOR DELAY to throw:
+     * "The Microsoft Distributed Transaction Coordinator (MS DTC) has cancelled the distributed transaction."
+     *
+     * This differs from PostgreSQL/MySQL/MariaDB where the XAConnectionLock
+     * read lock blocks end(TMFAIL) and the slow SQL completes normally.
+     */
+    @Override
+    boolean slowSQLThrowsOnReaperTimeout() {
+        return true;
+    }
+
+    @Override
+    String createTableDDL() {
+        return "IF OBJECT_ID('xa_reaper_test', 'U') IS NULL " +
+               "CREATE TABLE xa_reaper_test ( id INT PRIMARY KEY, val VARCHAR(100) )";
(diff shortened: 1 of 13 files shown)


agroal-narayana/src/main/java/io/agroal/narayana/BaseXAResource.java
```diff
@@ -66,6 +66,9 @@ public void commit(Xid xid, boolean onePhase) throws XAException {
     public void end(Xid xid, int flags) throws XAException {
         try {
             xaResource.end( xid, flags );
+            if ( ( flags & XAResource.TMFAIL ) != 0 ) {
+                transactionAware.transactionBeforeCompletion( false );
+            }
```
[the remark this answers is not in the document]
gastaldi: Right, I think that's better, let me push this change

gastaldi (approved):

marcosgopen:
> @marcosgopen: Is this MR now ready for review or still planed adding lock mechanism from earlier versions?

It is ready for review! Thanks @graben 
I believe the locking mechanism is not needed as the George's fix is enough to solve the issue.

marcosgopen:
@gastaldi  Should I squash and mention the AG-300 in the commit message?

gastaldi:
Sure, I think that makes sense

marcosgopen:
I have just removed the not needed byteman debug config for PostgreSQLTest which does not use byteman

graben:
> I have just removed the not needed byteman debug config for PostgreSQLTest which does not use byteman

Do I miss something. I still see byteman debug config and I think it is used?!

marcosgopen:
> > I have just removed the not needed byteman debug config for PostgreSQLTest which does not use byteman
> 
> Do I miss something. I still see byteman debug config and I think it is used?!

It is used in PostgreSQLXAReaperRaceTest, but not in https://github.com/agroal/agroal/blob/master/agroal-integration-tests/src/test/java/io/agroal/tests/PostgreSQLTest.java (it was added there by mistake)

agroal-integration-tests/src/test/java/io/agroal/tests/xa/MSSQLXAReaperRaceTest.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@Tag( "testcontainers" )
+@Testcontainers
+@BMUnitConfig(debug = true)
+class MSSQLXAReaperRaceTest extends XAReaperRaceTestBase {
+
+    private static final Logger logger = Logger.getLogger( MSSQLXAReaperRaceTest.class.getName() );
+
+    @Container
+    static MSSQLServerContainer mssql = new MSSQLServerContainer( "mcr.microsoft.com/mssql/server:2022-latest" )
```
marcosgopen: If I can review my change :) I would like to ask how the database images versions will be updated (dependabot?) and which versions should be tested and which not.
It might be too early to mention this, but please note that '2022-latest' for mssql might be unstable to test as an update in the image could possibly cause the test output to be different.
 cc @graben

graben: Good question!

graben (approved):
Successfully tested against narayana-spring-boot IT test suite

gastaldi:
@marcosgopen the agroal-integration-tests didn't exist in 2.7, so I think it would be better if the fix were in a separate commit so @barreiro can cherry-pick and backport to the older versions.

marcosgopen:
> @marcosgopen the agroal-integration-tests didn't exist in 2.7, so I think it would be better if the fix were in a separate commit so @barreiro can cherry-pick and backport to the older versions.

Good point, I will update the PR

.github/workflows/maven-action.yml
```diff
@@ -6,6 +6,10 @@ on:
   schedule:
     - cron: '0 5 * * 0'  # at 05:00 on Sunday
 
+concurrency:
```
gastaldi, line 9: This commit can be dropped, I moved to https://github.com/agroal/agroal/pull/181

agroal-integration-tests/src/test/java/io/agroal/tests/xa/MSSQLXAReaperRaceTest.java
gastaldi: I pushed https://github.com/agroal/agroal/pull/179/changes/b57b8d52b438e64a8440eef2fdde7054e18f6fb8 moving the image names (and tags) to the pom.xml. Renovate could be able to update it AFAIK, Dependabot will not work in this case

Here is an example of a `renovate.json` that would update it:

```json
{
  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
  "extends": [
    "config:recommended"
  ],
  "customManagers": [
    {
      "customType": "regex",
      "description": "Update Docker image versions in Maven POM properties",
      "fileMatch": [
        "(^|/)pom\\.xml$"
      ],
      "matchStrings": [
        "<[a-z.]+testcontainer\\.image>(?<depName>[^:]+):(?<currentValue>[^<]+)</[a-z.]+testcontainer\\.image>"
      ],
      "datasourceTemplate": "docker"
    }
  ]
}
```

marcosgopen: Thanks George!

gastaldi: I have also created https://redhat.atlassian.net/browse/AG-301 asking to enable Renovate in this repository. It should be able to create PRs like the following: https://github.com/gastaldi/agroal/pull/6

gastaldi:
I believe this is good to be merged. Any objections @barreiro ?

barreiro (approved):
thanks @gastaldi !! this looks good but needs to go as a new issue, not AG-300. this is a different problem.

gastaldi:
~@marcosgopen can you reword your commits so it points to `AG-304` instead?~

Already did, nvm ;)

gastaldi:
@barreiro seems like the repository requires signed commits, but `Rebase and merge` fails with `Base branch requires signed commits. Rebase merges cannot be automatically signed by GitHub`

graben:
> @barreiro seems like the repository requires signed commits, but `Rebase and merge` fails with `Base branch requires signed commits. Rebase merges cannot be automatically signed by GitHub` 

Did you try rebase on GitHub or local? Only later really works!

gastaldi:
@graben you're right, it only works locally. Thanks!

PR #182, opened 2026-03-30
AG-302 - Flush all idle connections on first validation failure to speed up recovery after DB failover

gastaldi:
When a database fails over without closing TCP connections (e.g. AWS RDS), every pooled connection becomes stale. Previously, each one was validated individually, blocking for the full socket timeout per connection. This made the application unavailable for (timeout × pool size) after failover.

Now, once the first connection fails validation, all remaining idle connections are flushed immediately without attempting isValid() on each, reducing recovery time from O(N × timeout) to O(1 × timeout). On-borrow validation also skips stale connections while the pool is being refreshed.

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -449,7 +449,7 @@ public final class ConnectionPool implements Pool {
     // handler must be in VALIDATION state
     private boolean performValidation(ConnectionHandler handler, boolean idle) {
         fireBeforeConnectionValidation( listeners, handler );
-        if ( handler.isValid() && idle ? handler.passValidationToIdle() : handler.passValidationToActive() ) {
+        if ( handler.isValid() && ( idle ? handler.passValidationToIdle() : handler.passValidationToActive() ) ) {
             fireOnConnectionValid( listeners, handler );
             if ( idle ) {
                 handlerTransferQueue.tryTransfer( handler );
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -106,6 +106,8 @@ public final class ConnectionPool implements Pool {
     private final LongAccumulator maxUsed = new LongAccumulator( Math::max, Long.MIN_VALUE );
     private final LongAdder activeCount = new LongAdder();
 
+    private volatile boolean stalePoolDetected;
+
     private MetricsRepository metricsRepository;
     private ConnectionCache localCache;
     private List<AgroalPoolInterceptor> interceptors;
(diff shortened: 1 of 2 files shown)


agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -851,25 +859,26 @@ private final class ValidationTask implements Runnable {
...
+                    if ( handler.tryValidationFromIdle() ) {
+                        if ( !performValidation( handler, true ) ) {
+                            failureDetected = true;
+                            stalePoolDetected = true;
+                        }
+                    }
                 }
             }
+            if ( failureDetected ) {
+                fill( configuration.minSize() );
```
graben: I would suggest du first switch flag back to false before filling pool

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -851,25 +859,26 @@ private final class ValidationTask implements Runnable {
         public void run() {
             housekeepingExecutor.schedule( this, configuration.validationTimeout().toNanos(), NANOSECONDS );
 
+            boolean failureDetected = false;
             for ( ConnectionHandler handler : allConnections ) {
-                housekeepingExecutor.execute( new ValidateConnectionTask( handler ) );
```
graben, line 855: Validation will be changed from multiple shortrunning tasks to one "big" task. Hard to calculate how this might behave in several szenarios.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
```diff
@@ -449,7 +457,7 @@ private boolean borrowValidation(ConnectionHandler handler) {
     // handler must be in VALIDATION state
     private boolean performValidation(ConnectionHandler handler, boolean idle) {
         fireBeforeConnectionValidation( listeners, handler );
-        if ( handler.isValid() && idle ? handler.passValidationToIdle() : handler.passValidationToActive() ) {
+        if ( handler.isValid() && ( idle ? handler.passValidationToIdle() : handler.passValidationToActive() ) ) {
```
graben, line 460: Looks like a bugfix either way?

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
gastaldi: Yeah, that's a valid concern, however the overhead should be modest, since the extra work is just `tryFlushFromIdle()` + `flushHandler()` calls (no `isValid()` network round-trips)

graben: Indeed, that's true.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
gastaldi: Yes, I made it a separate commit for clarity and backportability

graben (approved):

barreiro (changes requested):
can't simply call `flushPool(GRACEFULL)` ?? that way that logic was reused

otherwise, I would rather continue to have validation in individual pieces of work (tasks) than a single one

the idea is good :)

barreiro (approved):

PR #183, opened 2026-03-30
AG-303 - Replace dynamic proxies with static classes to minimize reflection

gastaldi:
Replace `Proxy.newProxyInstance()` / `InvocationHandler` based CLOSED sentinels with singleton implementations in all seven wrapper classes (`Connection`, `Statement`, `PreparedStatement`, `CallableStatement`, `ResultSet`, `XAConnection`, `XAResource`).

This eliminates all reflection from the wrapper package, enabling direct method dispatch instead of reflective method-name switching, better JIT inlining, and GraalVM/AOT compatibility.

Also cache the `setProperty` method lookup in `PropertyInjector` to avoid a per-call `cls.getMethod()` on the fallback injection path.

Fix `ClosedXAResource` to throw `XAException` with a descriptive message instead of the previous `SQLException` that was incompatible with the `XAResource` interface contract.


graben (approved):

graben:
@gastaldi : what do you think about extract new classes? Actually you implement statement interface multiple times. Maybe inheritance would make it less boiled?!

gastaldi:
Not a bad idea, will take a look

.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
```
graben: Intended checkin?

gastaldi: Oops, nope. Thanks for catching it

barreiro:
I second Benjamin's comment. 

can you move this closed classes to their own package and give them a format them more in line with the rest of the project ?? thanks!!

gastaldi:
> I second Benjamin's comment.
> 
> can you move this closed classes to their own package and give them a format them more in line with the rest of the project ?? thanks!!

Done

PR #184, opened 2026-04-14
AG-305 - Use Connection.setNetworkTimeout as a hard backstop in ConnectionValidator#sqlValidator

gastaldi:
`Statement.setQueryTimeout` relies on the driver sending a cancel signal to the server, which may not work when the connection is stuck at the network level (e.g. network partition, firewall dropping packets).

Setting a network timeout on the socket ensures validation will return within the configured timeout regardless of server cooperation.

### DIFF
--- a/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
+++ b/agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
@@ -11,6 +11,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.util.Collection;
+import java.util.concurrent.Executor;
 
 /**
  * The configuration of the connection pool.
@@ -211,6 +212,9 @@ public interface AgroalConnectionPoolConfiguration {
      */
     interface ConnectionValidator {
 
+        // extra buffer so setQueryTimeout can cancel the statement cleanly before the network timeout forces the socket closed
+        Duration NETWORK_TIMEOUT_BUFFER = Duration.ofSeconds( 5 );
+
         /**
          * The default validation strategy {@link Connection#isValid(int)}
          */
@@ -260,17 +264,58 @@ public interface AgroalConnectionPoolConfiguration {
          * A validator that uses the provided SQL statement for validation with a timeout (in seconds).
          * If the timeout period expires before the operation completes, the connection is invalidated.
          * A timeout of 0 means no timeout.
+         *
          */
         static ConnectionValidator sqlValidator(String sql, int timeoutSeconds) {
+            Duration duration = Duration.ofSeconds( timeoutSeconds );
+            return sqlValidator( sql, duration );
+        }
+
+        /**
+         * A validator that uses the provided SQL statement for validation with a query timeout.
+         * If the query timeout period expires before the operation completes, the connection is invalidated.
+         * A query timeout of zero means no timeout.
+         */
+        static ConnectionValidator sqlValidator(String sql, Duration queryTimeout) {
+            return sqlValidator( sql, queryTimeout, queryTimeout.isZero() ? null : queryTimeout.plus( NETWORK_TIMEOUT_BUFFER ) );
+        }
+
+        /**
+         * A validator that uses the provided SQL statement for validation with a query timeout and a configurable network timeout.
+         * If the timeout period expires before the operation completes, the connection is invalidated.
+         * A timeout of zero means no timeout.
+         * The network timeout should be greater than the query timeout to allow the query timeout to cancel the statement
+         * cleanly before the network timeout forces the socket closed.
+         * A null network timeout will keep the connection's network timeout configuration unchanged
+         */
+        static ConnectionValidator sqlValidator(String sql, Duration queryTimeout, Duration networkTimeout) {
             return new ConnectionValidator() {
                 @Override
                 public boolean isValid(Connection connection) {
+                    Executor executor = Runnable::run;
+                    int originalNetworkTimeout = 0;
+                    if ( networkTimeout != null ) {
+                        try {
+                            originalNetworkTimeout = connection.getNetworkTimeout();
+                            connection.setNetworkTimeout( executor, (int) networkTimeout.toMillis() );
+                        } catch ( Exception ignore ) {
+                            // driver may not support network timeout
+                        }
+                    }
                     try (var statement = connection.createStatement()) {
-                        statement.setQueryTimeout( timeoutSeconds );
+                        statement.setQueryTimeout( (int) queryTimeout.toSeconds() );
                         statement.execute( sql );
                         return true;
-                    } catch (Exception t) {
+                    } catch ( Exception t ) {
                         return false;
+                    } finally {
+                        if ( networkTimeout != null ) {
+                            try {
+                                connection.setNetworkTimeout( executor, originalNetworkTimeout );
+                            } catch ( Exception ignore ) {
+                                // driver may not support network timeout
+                            }
+                        }
                     }
                 }
             };
--- a/agroal-test/src/test/java/io/agroal/test/basic/PropertiesReaderTests.java
+++ b/agroal-test/src/test/java/io/agroal/test/basic/PropertiesReaderTests.java
@@ -18,6 +18,7 @@ import java.lang.reflect.Field;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.Connection;
+import java.time.Duration;
 import java.time.LocalTime;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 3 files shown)


agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -260,17 +260,35 @@ public boolean isValid(Connection connection) {
          * A validator that uses the provided SQL statement for validation with a timeout (in seconds).
          * If the timeout period expires before the operation completes, the connection is invalidated.
          * A timeout of 0 means no timeout.
+         * Uses {@link Connection#setNetworkTimeout(java.util.concurrent.Executor, int)} as a hard backstop
+         * in addition to {@link java.sql.Statement#setQueryTimeout(int)}, since the query cancellation signal
+         * may not work when the connection is stuck at the network level.
          */
         static ConnectionValidator sqlValidator(String sql, int timeoutSeconds) {
```
gsmet, line 269: And we can specify a specific timeout for the validation query? In a way that could be defined by a Quarkus config property?

gastaldi: We already have that with https://quarkus.io/guides/datasource#quarkus-agroal_quarkus-datasource-jdbc-validation-query-timeout

See: https://github.com/quarkusio/quarkus/blob/main/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSources.java#L345

gsmet: OK, then it's very much in line with what I suggested so I would say +1. But we definitely need Luis to verify it makes sense :)

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -260,17 +260,63 @@ public boolean isValid(Connection connection) {
...
+        static ConnectionValidator sqlValidator(String sql, Duration queryTimeout, Duration networkTimeout) {
             return new ConnectionValidator() {
                 @Override
                 public boolean isValid(Connection connection) {
+                    int originalNetworkTimeout = 0;
+                    try {
+                        originalNetworkTimeout = connection.getNetworkTimeout();
+                        if ( !networkTimeout.isZero() ) {
+                            connection.setNetworkTimeout( Runnable::run, (int) networkTimeout.toMillis() );
+                        }
```
gsmet: What I'm not entirely sure of is if it makes always sense to override the network timeout with the timeout value.

Typically, there might be good reasons why you set it high globally as establishing the connection might be slow.
Is there a default connection timeout in either Agroal or Quarkus?

gastaldi: This setting was introduced in https://github.com/agroal/agroal/commit/f1fa26be54810c040043db1e81ccbcdba4859422 and it's not in a released version yet

graben:
Hi @gastaldi, just thinking about AgroalPropertiesReader. Actually only validation timeout is supported. It's also integrated into Spring-Boot starter to configure validator. Might worth thinking about a backwards compatible enhancement! 🤔

gastaldi:
I've now introduced `io.agroal.api.configuration.AgroalConnectionPoolConfiguration.ConnectionValidator#sqlValidator(java.lang.String, java.time.Duration, java.time.Duration)` and kept the implementation   in the existing methods

gastaldi:
I can confirm that this method fixes https://github.com/quarkusio/quarkus/issues/49976, I'll submit a PR using it in Quarkus meanwhile

agroal-test/src/test/java/io/agroal/test/basic/PropertiesReaderTests.java
```diff
@@ -51,25 +50,6 @@ void basicPropertiesReaderTest() throws IOException {
...
-        Assertions.assertEquals( "io.agroal.api.configuration.AgroalConnectionPoolConfiguration$ConnectionValidator$4", validator.getClass().getName() );
-        Assertions.assertEquals( "select 1", getDeclaredField( validator, "val$sql" ) );
-        Assertions.assertEquals( 0, getDeclaredField( validator, "val$timeoutSeconds" ) );
-
-        connectionValidatorName = "sql[select 1]5000";
-        validator = AgroalPropertiesReader.parseConnectionValidator(connectionValidatorName);
-        Assertions.assertEquals( "io.agroal.api.configuration.AgroalConnectionPoolConfiguration$ConnectionValidator$4", validator.getClass().getName() );
-        Assertions.assertEquals( "select 1", getDeclaredField( validator, "val$sql" ) );
-        Assertions.assertEquals( 5000, getDeclaredField( validator, "val$timeoutSeconds" ) );
-    }
```
gastaldi, line 71: I removed this test because I believe we should be testing if the validator does what is supposed to do, not its internal structure

graben: Well, we are loosing the only test case that prove custom queries. Hmm... 🤔

barreiro: I agree, the test should remain to ensure the properties are read correctly

gastaldi: Agree to disagree :) This test is fragile, again, we should be testing the behavior instead. Anyway, I fixed the test to pass now.

agroal-test/src/test/java/io/agroal/test/basic/PropertiesReaderTests.java
```diff
@@ -61,13 +62,13 @@ void sqlValidatorTest() throws NoSuchFieldException, IllegalAccessException {
         validator = AgroalPropertiesReader.parseConnectionValidator(connectionValidatorName);
         Assertions.assertEquals( "io.agroal.api.configuration.AgroalConnectionPoolConfiguration$ConnectionValidator$4", validator.getClass().getName() );
         Assertions.assertEquals( "select 1", getDeclaredField( validator, "val$sql" ) );
-        Assertions.assertEquals( 0, getDeclaredField( validator, "val$timeoutSeconds" ) );
```
graben, line 64: @gastaldi: Isn't this this test case now testing something "different"? This was for method `static ConnectionValidator sqlValidator(String sql, int timeoutSeconds)` now it seems to test  `static ConnectionValidator sqlValidator(String sql, Duration queryTimeout)`?

gastaldi: Yeah, that's because if you look at the first method implementation, it just delegates the call to the later method

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -260,17 +264,53 @@ public boolean isValid(Connection connection) {
...
+         */
+        static ConnectionValidator sqlValidator(String sql, Duration queryTimeout, Duration networkTimeout) {
             return new ConnectionValidator() {
                 @Override
                 public boolean isValid(Connection connection) {
+                    Executor executor = Runnable::run;
+                    int originalNetworkTimeout = 0;
+                    try {
+                        originalNetworkTimeout = connection.getNetworkTimeout();
+                        connection.setNetworkTimeout( executor, (int) networkTimeout.toMillis() );
```
gastaldi: Should we always set the network duration or should we have a condition that the networkTimeout is kept unchanged?

gastaldi: Added a null check, if the parameter is null, then we keep it unchanged

agroal-api/src/main/java/io/agroal/api/configuration/AgroalConnectionPoolConfiguration.java
```diff
@@ -260,17 +264,58 @@ public boolean isValid(Connection connection) {
...
+            return sqlValidator( sql, duration );
+        }
+
+        /**
+         * A validator that uses the provided SQL statement for validation with a query timeout.
+         * If the query timeout period expires before the operation completes, the connection is invalidated.
+         * A query timeout of zero means no timeout.
+         */
+        static ConnectionValidator sqlValidator(String sql, Duration queryTimeout) {
+            return sqlValidator( sql, queryTimeout, queryTimeout.isZero() ? Duration.ZERO : queryTimeout.plus( NETWORK_TIMEOUT_BUFFER ) );
```
graben: Shouldn't networkTimeout be null at default?

gastaldi: Doh, good catch, fixed

graben (approved):

barreiro (approved):

PR #188, opened 2026-05-04
AG-304 - use var handle in wrappers

barreiro:
replaces direct field access to the wrapped JDBC objects with `VarHandle` using getAcquire/setRelease memory ordering

this provides proper visibility guarantees across threads without the overhead of full volatile semantics

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/wrapper/CallableStatementWrapper.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/wrapper/CallableStatementWrapper.java
@@ -8,6 +8,8 @@ import io.agroal.pool.wrapper.closed.ClosedCallableStatement;
 
 import java.io.InputStream;
 import java.io.Reader;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
 import java.math.BigDecimal;
 import java.net.URL;
 import java.sql.Array;
@@ -33,20 +35,36 @@ import java.util.Map;
  * @author <a href="lbarreiro@redhat.com">Luis Barreiro</a>
  * @author <a href="jesper.pedersen@redhat.com">Jesper Pedersen</a>
  */
+@SuppressWarnings( "resource" )
 public final class CallableStatementWrapper extends StatementWrapper implements CallableStatement {
+
+    private static final VarHandle WRAPPED;
+
+    static {
+        try {
+            WRAPPED = MethodHandles.lookup().findVarHandle( CallableStatementWrapper.class, "wrappedStatement", CallableStatement.class );
+        } catch ( NoSuchFieldException | IllegalAccessException e ) {
+            throw new ExceptionInInitializerError( e );
+        }
+    }
+
+    private CallableStatement wrappedCallableStatement() {
+        return (CallableStatement) WRAPPED.getAcquire( this );
+    }
 
     // --- //
 
+    @SuppressWarnings( "unused" )
     private CallableStatement wrappedStatement;
 
     public CallableStatementWrapper(ConnectionWrapper connectionWrapper, CallableStatement statement, boolean trackJdbcResources, AutoCloseableElement<StatementWrapper> head, boolean defaultHoldability) {
         super( connectionWrapper, statement, trackJdbcResources, head, defaultHoldability );
-        wrappedStatement = statement;
+        WRAPPED.setRelease( this, statement );
     }
 
     @Override
     public void close() throws SQLException {
-        wrappedStatement = ClosedCallableStatement.INSTANCE;
+        WRAPPED.setRelease( this, ClosedCallableStatement.INSTANCE );
         super.close();
     }
 
@@ -56,7 +74,7 @@ public final class CallableStatementWrapper extends StatementWrapper implements 
     public void registerOutParameter(int parameterIndex, int sqlType) throws SQLException {
         try {
             verifyEnlistment();
-            wrappedStatement.registerOutParameter( parameterIndex, sqlType );
+            wrappedCallableStatement().registerOutParameter( parameterIndex, sqlType );
         } catch ( SQLException se ) {
             connection.getHandler().setFlushOnly( se );
             throw se;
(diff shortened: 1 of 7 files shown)


gastaldi (commented):
Not sure if that's overkill, but you know better than me 😀

agroal-pool/src/main/java/io/agroal/pool/wrapper/ConnectionWrapper.java
```diff
@@ -151,8 +167,9 @@ public void close() throws SQLException {
     public void abort(Executor executor) throws SQLException {
         handler.traceConnectionOperation( "abort()" );
         try {
-            wrappedConnection = ClosedConnection.INSTANCE;
-            wrappedConnection.abort( executor );
+            Connection connection = wrappedConnection();
+            WRAPPED.setRelease( this, ClosedConnection.INSTANCE );
+            connection.abort( executor );
```
gastaldi, line 172: This is a bug fix, well done!

gastaldi (approved):

gastaldi:
Is the linked JIRA reference correct?

graben (approved):

gastaldi (commented):
`ConnectionWrapper#close()` has a TOCTOU (time-of-check-time-of-use) race:

```java
if ( wrappedConnection() != ClosedConnection.INSTANCE ) {   // getAcquire (read)
    WRAPPED.setRelease( this, ClosedConnection.INSTANCE );   // setRelease (write)
```

The read and write are two separate operations. If two threads (e.g., the application thread and the Narayana reaper) both call `close()` concurrently, both can see the non-closed value from `getAcquire`, both enter the `if` block, and both execute the cleanup code — `pruneClosed()`, `closeAllAutocloseableElements()`, `onConnectionWrapperClose()`.

Consider using an atomic swap instead:

```java
if ( WRAPPED.getAndSet( this, ClosedConnection.INSTANCE ) != ClosedConnection.INSTANCE ) {
    pruneClosed();
    if ( trackedStatements != null ) {
        addLeakedStatements( trackedStatements.closeAllAutocloseableElements() );
    }
    handler.onConnectionWrapperClose( this );
}
```

`getAndSet` atomically reads the old value and writes the sentinel in one shot. Only the thread that swaps away the real connection enters the block — the other sees `ClosedConnection.INSTANCE` and skips it.

barreiro:
> Is the linked JIRA reference correct?

it addresses the (potential) interleaving issue. I can feel a new issue specifically for this if you want,

gastaldi:
No need for a new JIRA issue, it's fine to reuse the same if it's related

gastaldi (commented):
## Audit of all `close()` methods — TOCTOU and double-read issues

Since the goal of AG-304 is to fix interleaving with the Narayana reaper thread, all `close()` methods that guard on the sentinel should use atomic swaps to prevent concurrent threads from executing cleanup code twice.

### `XAConnectionWrapper#close()` — same TOCTOU as `ConnectionWrapper#close()`

```java
if ( wrappedXAConnection() != ClosedXAConnection.INSTANCE ) {   // getAcquire
    WRAPPED.setRelease( this, ClosedXAConnection.INSTANCE );     // setRelease
    // cleanup: closeAllAutocloseableElements, handler.onConnectionWrapperClose
}
```

Same race as `ConnectionWrapper#close()` — two threads can both pass the guard and run cleanup twice. Same fix applies:

```java
if ( WRAPPED.getAndSet( this, ClosedXAConnection.INSTANCE ) != ClosedXAConnection.INSTANCE ) {
    ...
}
```

### `StatementWrapper#close()` — double-read

```java
if ( wrappedStatement() != ClosedStatement.INSTANCE ) {  // 1st getAcquire
    ...
    wrappedStatement().close();                           // 2nd getAcquire
}
```

Between the two `getAcquire` calls, another thread could swap the field to the sentinel. The guard passes on the real statement, but `.close()` is then called on `ClosedStatement.INSTANCE`. Likely benign (the sentinel's `close()` is probably a no-op), but it's a logical race. Should cache the read in a local:

```java
Statement stmt = wrappedStatement();
if ( stmt != ClosedStatement.INSTANCE ) {
    ...
    stmt.close();
}
```

### `XAResourceWrapper#close()` — TOCTOU but harmless

```java
if ( wrappedXAResource() != ClosedXAResource.INSTANCE ) {
    WRAPPED.setRelease( this, ClosedXAResource.INSTANCE );
}
```

Two threads could both enter, but the body is just the idempotent `setRelease` — no side-effectful cleanup. Not a bug, but `getAndSet` would be consistent with the other wrappers.

### The rest are fine

- **`CallableStatementWrapper#close()` / `PreparedStatementWrapper#close()`**: unconditional `setRelease` + `super.close()`, no guard, no TOCTOU.
- **`ResultSetWrapper#close()`**: no guard, always calls `close()` on the current value, sentinel set in `finally`. Same behavior as the original.
- **`ConnectionWrapper#abort()`**: correctly caches the real connection in a local before setting the sentinel — this is actually a bugfix over the original code.

barreiro:
@gastaldi the cleanup code that is executed is idempotent (it's safe to be called multiple times) and should also be safe to be called concurrently.

I would like to avoid the overhead of a full `getAndSet` operation. the chance of ending up executing the cleanup code twice is low.

agroal-pool/src/main/java/io/agroal/pool/wrapper/CallableStatementWrapper.java
```diff
@@ -33,20 +35,36 @@
...
 
     public CallableStatementWrapper(ConnectionWrapper connectionWrapper, CallableStatement statement, boolean trackJdbcResources, AutoCloseableElement<StatementWrapper> head, boolean defaultHoldability) {
         super( connectionWrapper, statement, trackJdbcResources, head, defaultHoldability );
-        wrappedStatement = statement;
+        WRAPPED.setRelease( this, statement );
     }
 
     @Override
     public void close() throws SQLException {
-        wrappedStatement = ClosedCallableStatement.INSTANCE;
```
franz1981, line 49: mmm this was totally unguarded before?

franz1981 (changes requested):
If the previous way to write/read the field was unguarded - it means it was relying on external synchronization to make it visible (and read it too). 
If it was correct before, it means you can keep it as it was. If it was wrong (because it proved to be wrong!) - than acquire/release is correct as you have done it here

PR #190, opened 2026-05-14
AG-308 - Apply setTransactionIsolation to all connections, not just the first

gastaldi:
- #165 moved setTransactionIsolation inside the double-checked locking block in loadDefaults() as part of the lazy initialization of defaults. This inadvertently turned a per-connection call into a one-time-only call — only the first connection got the configured isolation level, while all subsequent connections kept the driver default.


gastaldi:
@barreiro we need a release with this fix ASAP, see https://github.com/quarkusio/quarkus/issues/54243

graben (approved):

Issue #192, opened 2026-05-28
Logical error in ConnectionWrapper.hasTrackedStatements()


Siggen:
Hello,

https://github.com/agroal/agroal/blob/01176eeb2182702f2b265efaa115953f7120e5bf/agroal-pool/src/main/java/io/agroal/pool/wrapper/ConnectionWrapper.java#L147

I stumbled on the following code in `ConnectionWrapper` and the logic seems wrong: 

    public boolean hasTrackedStatements() throws SQLException {
        return trackedStatements != null && trackedStatements.isElementListEmpty();
    }

If `trackedStatements` list is empty, why would `hasTrackedStatements` return `true` ?

I would propose following implementation:

    public boolean hasTrackedStatements() throws SQLException {
        return trackedStatements != null && !trackedStatements.isElementListEmpty();
    }

Thanks!

gastaldi:
Moved to https://redhat.atlassian.net/browse/AG-310

PR #193, opened 2026-05-29
AG-310 - Fix inverted logic in 'hasTrackedStatements' method.

Siggen:
Fix for #192.


graben:
@Siggen: Just see you did the same the same time :)

graben:
@barreiro, @gastaldi : Seems to be a cosmetic error only since double negation keeps the flow on a valid path.

Siggen:
> Seems to be a cosmetic error only since double negation keeps the flow on a valid path.

Yes, you are right: there is no practical consequences. Everything is working fine.

It is more semantic error where a method does the opposite of what it is meant to do :-)

gastaldi:
Can you please sign your commit? That's one thing we require in our PRs

Siggen:
> Can you please sign your commit? That's one thing we require in our PRs

Of course! It is done (hopefully).

Siggen:
I mangled the commit message and I had to force push again (sorry for the noise).

Siggen:
Thanks a lot!

PR #200, opened 2026-06-08
AG-314 - Fix XA timeout exception loses SQLState when TMFAIL closes wrappers before end()

francisco-bru:
- Fixes https://redhat.atlassian.net/browse/AG-314

## Problem

When the Narayana reaper thread fires a transaction timeout and calls `XAResource.end(TMFAIL)`, `transactionBeforeCompletion(false)` was being called **before** `xaResource.end()`. This closed all `ConnectionWrapper` instances (replacing the underlying connection with `ClosedConnection.INSTANCE`) before the JDBC driver's cancellation exception could propagate cleanly.

As a result, the `MySQLStatementCancelledException` (SQLState `70100`) surfaced as a generic `SQLException` with no SQLState, causing Spring's `SQLExceptionTranslator` to produce `UncategorizedSQLException` instead of `QueryTimeoutException`.

## Fix

Move `transactionBeforeCompletion(false)` to **after** `xaResource.end()` — both on the success path and in the `catch (XAException)` block — so the driver exception propagates with its original SQLState intact.

```java
// Before (problematic order)
if ( ( flags & XAResource.TMFAIL ) != 0 ) {
    transactionAware.transactionBeforeCompletion( false );  // closes wrappers FIRST
}
xaResource.end( xid, flags );

// After (fixed order)
xaResource.end( xid, flags );  // driver exception propagates cleanly
if ( ( flags & XAResource.TMFAIL ) != 0 ) {
    transactionAware.transactionBeforeCompletion( false );
}
```

## Testing

All 36 existing Narayana tests pass with this change:
- `ReapTests` (2)
- `CompletionExceptionTests` (2)
- `ExceptionTests` (4)
- `MultipleWrappersTests` (2)
- `HoldOnCompletionTests` (20)

### DIFF
--- a/agroal-integration-tests/src/test/java/io/agroal/tests/xa/MySQLXASQLStateTest.java
+++ b/agroal-integration-tests/src/test/java/io/agroal/tests/xa/MySQLXASQLStateTest.java
@@ -1,0 +1,108 @@
+package io.agroal.tests.xa;
+
+import io.agroal.api.AgroalDataSource;
+import io.agroal.tests.Datasources;
+import jakarta.transaction.TransactionManager;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.mysql.MySQLContainer;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.logging.Logger;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/**
+ * AG-314: Verifies that when a transaction times out while a query is blocked,
+ * the resulting SQLException preserves the driver's original SQLState.
+ *
+ * Uses row-level lock contention: a raw JDBC connection holds an exclusive lock
+ * on a row, and the XA connection tries to acquire the same lock. When the
+ * Narayana reaper fires TMFAIL, the blocked query must be interrupted with an
+ * exception that carries a non-null SQLState.
+ */
+@Tag( "testcontainers" )
+@Testcontainers
+class MySQLXASQLStateTest {
+
+    private static final Logger logger = Logger.getLogger( MySQLXASQLStateTest.class.getName() );
+
+    @Container
+    static MySQLContainer mysql = new MySQLContainer( System.getProperty( "mysql.testcontainer.image", "mysql:8.0" ) );
+
+    private void verifyXASupported( AgroalDataSource dataSource ) {
+        TransactionManager txManager = com.arjuna.ats.jta.TransactionManager.transactionManager();
+        try {
+            txManager.setTransactionTimeout( 10 );
+            txManager.begin();
+            dataSource.getConnection().close();
+            txManager.rollback();
+        } catch ( Exception e ) {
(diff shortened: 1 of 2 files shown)


gastaldi (dismissed):
LGTM, any objections @marcosgopen?

gastaldi:
@francisco-bru any chance to provide a test that reproduces this?

gastaldi:
Investigation found that the null SQLState is a **MySQL Connector/J 9.x driver bug**, not an Agroal issue.

When `Statement.cancel()` triggers `KILL QUERY`, the driver creates a `MySQLStatementCancelledException` without setting the SQLState (it was `"70100"` in Connector/J 8.x). This causes Spring's `SQLExceptionTranslator` to produce `UncategorizedSQLException` instead of a properly categorized exception.

The ordering fix in this PR is independently correct — `transactionBeforeCompletion(false)` should run after `xaResource.end()`, not before. I've also pushed:
- A gap fix for the `catch (Exception e)` path (was missing `transactionBeforeCompletion(false)`)
- An integration test (`MySQLXASQLStateTest`) that reproduces the issue using lock contention — currently expected to fail until the driver bug is fixed upstream

gastaldi:
After deeper analysis, this PR should not be merged. Here's why:

**The reordering is counterproductive for TMFAIL on a busy connection.**

The original ordering (cleanup before `xaResource.end()`) is correct:
1. `transactionBeforeCompletion(false)` cancels the running query via `KILL QUERY`, freeing the connection
2. `xaResource.end(TMFAIL)` can then send `XA END` because the connection is no longer busy

The PR's reversed ordering tries to send `XA END` on a busy connection first — MySQL's protocol is synchronous, so this blocks or throws. The cleanup then runs in the catch block anyway, gaining nothing.

**The null SQLState is a MySQL Connector/J 9.x driver bug, not an Agroal bug.**

`MySQLStatementCancelledException` is created without setting the SQLState (it was `"70100"` in 8.x). The ordering of `transactionBeforeCompletion` vs `xaResource.end()` does not affect how the driver constructs the exception.

Filed upstream: https://github.com/mysql/mysql-connector-j/issues/125

gastaldi:
**Correction to my earlier comment:** The null SQLState is NOT a regression from 8.x — the same code path exists in both Connector/J 8.x and 9.x. `SQLExceptionsMapping.translateException()` has always used `new MySQLStatementCancelledException(ex.getMessage())` without passing the SQLState. The bug has been there all along.

See https://github.com/mysql/mysql-connector-j/issues/125 and https://github.com/mysql/mysql-connector-j/pull/126.

marcosgopen:
> After deeper analysis, this PR should not be merged. Here's why:
> 
> **The reordering is counterproductive for TMFAIL on a busy connection.**
> 
> The original ordering (cleanup before `xaResource.end()`) is correct:
> 
>     1. `transactionBeforeCompletion(false)` cancels the running query via `KILL QUERY`, freeing the connection
> 
>     2. `xaResource.end(TMFAIL)` can then send `XA END` because the connection is no longer busy
> 
> 
> The PR's reversed ordering tries to send `XA END` on a busy connection first — MySQL's protocol is synchronous, so this blocks or throws. The cleanup then runs in the catch block anyway, gaining nothing.
> 
> **The null SQLState is a MySQL Connector/J 9.x driver bug, not an Agroal bug.**
> 
> `MySQLStatementCancelledException` is created without setting the SQLState (it was `"70100"` in 8.x). The ordering of `transactionBeforeCompletion` vs `xaResource.end()` does not affect how the driver constructs the exception.
> 
> Filed upstream: [mysql/mysql-connector-j#125](https://github.com/mysql/mysql-connector-j/issues/125)

Thanks @gastaldi for the deep analysis. I agree this is not an Agroal issue. If we have a test to reproduce the reported issue I think it is a good chance to add some extra integration tests to verify all other db drivers work fine.

PR #202, opened 2026-06-22
AG-315 - Offload JDBC connection creation to platform threads to avoid virtual thread pinning

### LINKED ISSUE #201: Collaborative (inline) connection creation in handlerFromSharedCache causes virtual thread pinning and zombie transactions
## Problem

AG-290 changed `ConnectionPool.handlerFromSharedCache()` to create connections **inline on the calling thread** ("collaboratively") instead of delegating to the housekeeping executor. This causes a regression when used with Java 21+ virtual threads.

## Root cause

When a virtual thread calls `handlerFromSharedCache()` and no pooled connection is available, it now calls `createAndPoolConnection()` directly. This performs TCP connect, SSL handshake, and JDBC authentication on the virtual thread itself. Most JDBC drivers (PostgreSQL, Oracle, etc.) use `synchronized` blocks internally during connection establishment, which **pins** the virtual thread to its carrier thread.

Under load with many concurrent virtual threads, this leads to:
1. **Carrier thread starvation**: all carrier threads are pinned by virtual threads creating JDBC connections
2. **Transaction timeouts**: other virtual threads cannot make progress because no carrier threads are available
3. **Zombie transactions**: Narayana's Transaction Reaper intervenes on timed-out transactions, but the pinned virtual threads are still associated, causing `ARJUNA012381: Action completed with multiple threads` warnings and zombie transaction markers
4. **Cascading failure**: event loop thread blocking, high CPU, service unresponsiveness

## Why this didn't happen before AG-290

In the previous implementation, `handlerFromSharedCache()` submitted a `CreateConnectionTask` to the **housekeeping executor** and waited on `Future.get()`. The virtual thread blocked on `Future.get()`, which is virtual-thread-friendly (it parks the virtual thread without pinning the carrier). The carrier thread was free to run other virtual threads while waiting.

## Reproduction

1. Quarkus 3.36.x app with `quarkus-jdbc-postgresql` and `@RunOnVirtualThread` endpoints
2. Run under load in a resource-constrained environment (e.g., 0.2 vCPU container)
3. Observe: zombie transaction warnings, event loop blocking, CPU spikes
4. Same app with platform threads: no issue
5. Same app on Quarkus 3.35.x (Agroal 3.0): no issue

Symptoms reported by users:
```
ARJUNA012117: TransactionReaper::check processing TX in state SCHEDULE_CANCEL
ARJUNA012095: Abort of action invoked while multiple threads active within it
ARJUNA012381: Action completed with multiple threads - thread quarkus-virtual-thread-X was in progress with java.base/jdk.internal.vm.Continuation.doYield
ARJUNA012120: TransactionReaper::check worker not responding to interrupt -- worker marked as zombie
```

## Proposed fix

Revert `handlerFromSharedCache()` to always delegate connection creation to the housekeeping executor instead of creating inline. The calling thread is blocked waiting either way, so the inline approach provides negligible benefit while introducing virtual thread pinning risk.

The "collaborative" strategy could be made opt-in via configuration if there are workloads that benefit from it.

## Related

- Quarkus issue: https://github.com/quarkusio/quarkus/issues/54955
- User report: https://github.com/eggy03/PaperTrail-API-Quarkus/pull/3
- AG-304 (Narayana reaper interleaving fix) amplifies the impact — after `XAResource.end(TMFAIL)`, connection wrappers are poisoned via `transactionBeforeCompletion(false)`, causing exceptions in still-active virtual threads

gastaldi:
## Summary

- Offload `connectionFactory.createConnection()` to a dedicated cached platform thread pool to prevent virtual thread pinning on JDBC driver `synchronized` blocks
- Calling thread waits uninterruptibly on `Future.get()`, preserving identical blocking semantics to inline creation
- Parallel creation is maintained since the cached pool grows on demand

## Problem

AG-290 introduced inline ("collaborative") connection creation where the calling thread executes the JDBC driver's `connect()` call directly. Most JDBC drivers (PostgreSQL, Oracle, MySQL, etc.) use `synchronized` blocks during connection establishment, which pins virtual threads to their carrier threads. Under load with many virtual threads, this causes carrier thread starvation, transaction timeouts, and zombie transactions.

## Fix

Instead of changing _where_ `createAndPoolConnection()` is called, this changes _how_ the JDBC call runs within it. A `connectionCreationExecutor` (cached platform thread pool) runs `connectionFactory.createConnection()`, while the calling thread waits uninterruptibly for the result. Everything else in `createAndPoolConnection()` (permits, metrics, listeners, pool bookkeeping) continues to run on the calling thread.

## Test plan

- All 170 existing tests pass with zero modifications
- Existing `collaborativeConnectionCreationTest` validates parallel creation still works
- Existing `NewConnectionTimeoutTests` validate blocking semantics are preserved

Fixes https://github.com/agroal/agroal/issues/201

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -715,28 +715,21 @@ public final class ConnectionPool implements Pool {
      */
     private XAConnection createConnectionOnPlatformThread() throws SQLException {
         Future<XAConnection> future = connectionCreationExecutor.submit( connectionFactory::createConnection );
-        boolean interrupted = false;
         try {
-            while ( true ) {
-                try {
-                    return future.get();
-                } catch ( InterruptedException e ) {
-                    interrupted = true;
-                } catch ( ExecutionException e ) {
-                    if ( e.getCause() instanceof SQLException se ) {
-                        throw se;
-                    } else if ( e.getCause() instanceof RuntimeException re ) {
-                        throw re;
-                    } else if ( e.getCause() instanceof Error er ) {
-                        throw er;
-                    }
-                    throw new SQLException( "Exception while creating new connection", e.getCause() );
-                }
+            return future.get();
+        } catch ( InterruptedException e ) {
+            future.cancel( true );
+            currentThread().interrupt();
+            throw new SQLException( "Connection creation interrupted", e );
+        } catch ( ExecutionException e ) {
+            if ( e.getCause() instanceof SQLException se ) {
+                throw se;
+            } else if ( e.getCause() instanceof RuntimeException re ) {
+                throw re;
+            } else if ( e.getCause() instanceof Error er ) {
+                throw er;
             }
-        } finally {
-            if ( interrupted ) {
-                currentThread().interrupt();
-            }
+            throw new SQLException( "Exception while creating new connection", e.getCause() );
         }
     }
 
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -21,6 +21,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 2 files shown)


gastaldi:
This doesn't seem to fix the issue. Closing

PR #203, opened 2026-06-25
AG-315 - Disable collaborative connection creation on virtual threads

gastaldi:
## Summary

The AG-290 collaborative connection creation (`collaborate=true`) runs `createAndPoolConnection()` on the calling thread. On virtual threads this pins the carrier in the `handlerFromSharedCache()` CAS loop, causing CPU spikes and livelock under pool contention.

- On JDK 21+, set `collaborate=false` when the calling thread is virtual, offloading connection creation to the housekeeping executor
- On JDK < 21, behavior is unchanged (`collaborate=true`)
- Uses a multi-release JAR (`VirtualThreadUtil`) to detect virtual threads without reflection

## Test plan

- [x] All 30 Agroal tests pass
- [x] Verified with [Quarkus reproducer](https://github.com/quarkusio/quarkus/issues/54955) — health check responds in ~60ms (was timeout), no livelock
- [x] Verified with [PaperTrail reproducer](https://github.com/eggy03/PaperTrail-API-Quarkus/pull/3) in Docker with 0.15 CPU — zero zombie transactions, zero ARJUNA warnings
- [x] Verified that `collaborate=false` (pre-AG-290 behavior) eliminates the livelock, confirming AG-290 as root cause

Fixes https://redhat.atlassian.net/browse/AG-315

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
@@ -12,6 +12,7 @@ import io.agroal.api.transaction.TransactionIntegration;
 import io.agroal.pool.MetricsRepository.EmptyMetricsRepository;
 import io.agroal.pool.util.PriorityScheduledExecutor;
 import io.agroal.pool.util.StampedCopyOnWriteArrayList;
+import io.agroal.pool.util.VirtualThreadUtil;
 import io.agroal.pool.util.XAConnectionAdaptor;
 
 import javax.sql.XAConnection;
@@ -347,7 +348,7 @@ public final class ConnectionPool implements Pool {
     private ConnectionHandler handlerFromSharedCache() throws SQLException {
         long acquisitionTimeout = configuration.acquisitionTimeout().isZero() ? MAX_VALUE : configuration.acquisitionTimeout().toNanos();
         long deadline = acquisitionTimeout == MAX_VALUE ? MAX_VALUE : nanoTime() + acquisitionTimeout;
-        boolean collaborate = true;
+        boolean collaborate = !VirtualThreadUtil.isVirtualThread();
         int retries = configuration.establishmentRetryAttempts();
         try {
             for ( int i = 0; ; i++ ) {
--- a/agroal-pool/src/main/java/io/agroal/pool/util/VirtualThreadUtil.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/util/VirtualThreadUtil.java
@@ -1,0 +1,18 @@
+// Copyright (C) 2017 Red Hat, Inc. and individual contributors as indicated by the @author tags.
+// You may not use this file except in compliance with the Apache License, Version 2.0.
+
+package io.agroal.pool.util;
+
+/**
+ * Utility for virtual thread detection. On JDK < 21 this always returns false.
+ * On JDK 21+ the multi-release JAR overrides this with Thread.isVirtual().
+ */
+public final class VirtualThreadUtil {
+
+    private VirtualThreadUtil() {
+    }
+
+    public static boolean isVirtualThread() {
+        return false;
+    }
+}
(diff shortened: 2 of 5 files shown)


gastaldi:
For the record, Poolless always creates a new connection so it is not affected by this change and the error doesn't happen (when using `quarkus.datasource.jdbc.pooling-enabled=false`)

pavelkryl:
Built `fix-collaborate-vthread-pinning` locally (`3.3-SNAPSHOT`) and ran it against the [reproducer](https://gist.github.com/pavelkryl/b987aeadb89d41e5c13088085e666c4c) on **OpenJDK 25, `-Djdk.virtualThreadScheduler.parallelism=2`, `jdbc.max-size=2`, `ab -c 200`** (Quarkus 3.36.2):

| | Agroal 3.2 | this PR (3.3-SNAPSHOT) |
|---|---|---|
| process CPU under load | ~200% (both carriers pinned) | **~5%, flat** |
| requests completed | 0–1, `ab` times out | **196 / 196, 0 failed** |
| surplus virtual threads | spinning the carriers | **parked** (`agroal_awaiting_count=196`, `active=2`) |

Confirmed — the carrier busy-spin/livelock is gone; surplus borrowers now park cleanly on the no-connection path and drain at the pool's natural rate. Thanks for the quick turnaround! 🎉

agroal-pool/src/main/java/io/agroal/pool/util/VirtualThreadUtil.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private VirtualThreadUtil() {
+    }
+
+    public static boolean isVirtualThread() {
+        return false;
+    }
+
+    public static ExecutorService newVirtualThreadPerTaskExecutor() {
+        return null;
```
graben, line 22: Wouldn't throw UnsupportedOperationException be better?

barreiro (approved):
it's disappointing that this fix is needed, but not all the drivers are free from pins.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
graben, line 1: Btw there is a missing 'e' in line 380 'Connection'

agroal-test/src/test/java/io/agroal/test/basic/VirtualThreadTests.java
graben, line 1: Isn't 5000ms too much to really test that thread pinning is no problem at creation. We have 20 times 50ms and 2 times 200. So 2000ms should be fine? What do you think? Btw maybe also have a negative test scenario for collaborate true which shouldn't happen any more?

graben (changes requested):

graben:
@gastaldi : Just see @barreiro was faster.

agroal-pool/src/main/java/io/agroal/pool/ConnectionPool.java
barreiro: fixed. thanks for spotting !!

agroal-pool/src/main/java/io/agroal/pool/util/VirtualThreadUtil.java
gastaldi: Good idea, will do that in a separate PR

gastaldi: Actually, no, that will fail the test

barreiro: I'm fixing it

agroal-test/src/test/java/io/agroal/test/basic/VirtualThreadTests.java
barreiro: I changed the acquisition timeout to 2000ms and the latch timeout to 1000ms more

gsmet:
> Uses a multi-release JAR (VirtualThreadUtil) to detect virtual threads without reflection

Thanks for this extra effort, very much appreciated :).

PR #207, opened 2026-08-21
AG-317 - Support Request Boundaries

### LINKED ISSUE #206: Add support for Request Boundaries (beginRequest / endRequest)
The ConnectionWrapper currently does not implement the methods `beginRequest` and `endRequest` and therefore these methods fallback on the default method of the interface which is empty.

Now I know not every Implementation supports these calls, but maybe it's possible to activate it via config similar to how HikariCp does it?
We need to support TAC (Transparent Application Continuity) for Oracle and it is recommended to explicitly call beginRequest and endRequest.

Would it be possible to implement these Methods? Maybe hide it behind a flag so it only calls the methods on the underlying connection if it's explicitly enabled.

gastaldi:
## What

Adds support for the JDBC 4.3 Request Boundaries API — `Connection.beginRequest()` and `Connection.endRequest()` — in Agroal.

Previously these methods were not implemented and fell back to the empty default methods of the `Connection` interface, so features that rely on explicit request boundaries — notably Oracle's **Transparent Application Continuity (TAC)**, which recommends explicitly calling `beginRequest`/`endRequest` — could not work through Agroal.

The change is delivered as two commits:

### 1. Implement Request Boundaries in `ConnectionWrapper`

- `ConnectionWrapper` implements `beginRequest()` / `endRequest()`, delegating to the underlying connection with the same trace / flush-only idiom as the other wrapped methods. (No `verifyEnlistment()` guard — request boundaries are transaction-agnostic session-lifecycle hints, not transactional operations.)
- `ClosedConnection` throws `SQLException` on both methods, so invoking them on a closed wrapper behaves consistently with the rest of the connection API instead of silently falling through to the interface no-ops.

This lets an application or framework explicitly bracket its work.

### 2. Drive Request Boundaries from the pool on checkout/return

Per the JDBC specification, the pooling manager is the intended caller of these methods:

> The pooling manager should call `beginRequest` on the underlying `Connection` when the application obtains it from the pool, and `endRequest` when the application returns it.

- `ConnectionHandler` gains `beginRequest()` / `endRequest()` delegating to the raw connection (with flush-only handling on error).
- `ConnectionPool.afterAcquire()` calls `beginRequest()` on checkout.
- `ConnectionPool.returnConnectionHandler()` calls `endRequest()` right before `resetConnection()` — reached only after `transactionIntegration.disassociate()` succeeds, so enlisted/XA connections defer the call until the transaction actually completes.
- `PoolRequestBoundariesTests` covers checkout/return and per-borrow begin/end pairing.

With this, Oracle TAC works transparently without the application having to bracket its work.

## Notes

The original request suggested hiding this behind a HikariCP-style opt-in configuration flag. Since `Connection.beginRequest()`/`endRequest()` have default no-op implementations in the JDBC interface, the calls are safe for drivers that do not override them, so no flag is added. Happy to introduce an opt-in flag if preferred.

## Verification

Full build clean; `PoolRequestBoundariesTests` and `BasicNarayanaTests` (transaction path) all pass.

- Fixes #206
- Jira: https://redhat.atlassian.net/browse/AG-317

### DIFF
--- a/agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java
+++ b/agroal-pool/src/main/java/io/agroal/pool/ConnectionHandler.java
@@ -142,6 +142,29 @@ public final class ConnectionHandler implements TransactionAware, Acquirable {
         return xaResource;
     }
 
+    // --- Request Boundaries (JDBC 4.3) --- //
+
+    // Called by the pool when the connection is handed out, marking the start of a request.
+    // Per the JDBC contract a duplicate beginRequest() without an intervening endRequest() is ignored by the driver.
+    public void beginRequest() throws SQLException {
+        try {
+            connection.beginRequest();
+        } catch ( SQLException se ) {
+            setFlushOnly( se );
+            throw se;
+        }
+    }
+
+    // Called by the pool when the connection is returned, marking the end of a request.
+    public void endRequest() throws SQLException {
+        try {
+            connection.endRequest();
+        } catch ( SQLException se ) {
+            setFlushOnly( se );
+            throw se;
+        }
+    }
+
     @SuppressWarnings( "MagicConstant" )
     public void resetConnection() throws SQLException {
         transactionActiveCheck = NO_ACTIVE_TRANSACTION;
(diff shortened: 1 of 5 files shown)


graben (approved):
Looks good to me :)

PR #215, opened 2026-09-12
AG-320 - Do not busy-spin while waiting for a connection being created

### LINKED ISSUE #214: Busy-spin in ConnectionPool.handlerFromSharedCache when the pool is below max-size and connection creation is slow (not covered by #203)
Follow-up to #201, which was fixed by #203 ("AG-315 – Disable collaborative connection creation on virtual threads") and released in `agroal-pool` 3.2.1. That fix works; this is a second, distinct spin on the same method that it does not cover.

**Reproducer: https://github.com/pavelkryl/agroal-slow-connect-repro** (Maven, one class, dockerised Postgres, includes a negative control and the CPU profile).

## Summary

On **agroal-pool 3.2.1** — i.e. *with* #203 — a `@RunOnVirtualThread` workload still busy-spins its carrier threads in `ConnectionPool.handlerFromSharedCache`, provided that

1. the pool is **below `max-size`** (so Agroal can still grow it), **and**
2. **creating a new connection is slow** (a degraded or restarting database, a proxy queueing connections, a slow TLS handshake).

\#203 fixed the case where the pool is **at** `max-size`: there, threads park correctly and CPU stays flat. We verified that separately. But when the pool can still grow, the acquire loop appears to keep rescanning `allConnections` instead of waiting for the in-flight creation to complete — and with carriers == cores the process saturates every core.

We hit this in production: a managed-PostgreSQL restart made connection creation slow for a few minutes; the application then stayed pegged at 100% on both cores for **7 hours** and never recovered on its own. Restarting the service relapsed within ~40 seconds — a cold pool has to grow, which puts it straight back into this path. The only thing that broke the cycle was raising `-Djdk.virtualThreadScheduler.parallelism` well above the core count.

## Affected

- `agroal-pool` **3.2.1** (via Quarkus 3.37.4), JDK 25.0.3
- Reproduced on the JVM. Our production system runs a GraalVM native image and shows an identical signature.

## Reproducer

Full project: **https://github.com/pavelkryl/agroal-slow-connect-repro**

A single `@RunOnVirtualThread` `@Transactional` endpoint:

```java
@Path("/pool")
@RunOnVirtualThread
public class PoolResource {

    @Inject
    AgroalDataSource ds;

    @GET
    @Transactional
    public String hit() throws Exception {
        try (var c = ds.getConnection();
             var st = c.prepareStatement("select pg_sleep(0.5)")) {
            st.execute();
        }
        return "ok";
    }
}
```

```properties
quarkus.datasource.jdbc.max-size=30
quarkus.datasource.jdbc.min-size=2
quarkus.datasource.jdbc.acquisition-timeout=10
```

Run with `-Djdk.virtualThreadScheduler.parallelism=2` and load with `ab -c 300`.

**The trigger is a single server-side knob.** PostgreSQL's `post_auth_delay` makes every *new* backend sleep N seconds after authentication while leaving *existing* connections untouched — which is exactly what a database restart looks like to a connection pool:

```sql
ALTER SYSTEM SET post_auth_delay = '5s';
SELECT pg_reload_conf();
```

## Observed

With `post_auth_delay = 5s` and a pool starting cold:

```
 t(s)   active    await  maxUsed  available     CPU%
   10        1        0        1          0      206
   15        2        0        2          0      207
   20        3        0        3          0      202
   25        4        0        4          0      200
   ...
   70       13        0       13          0      201
```

- **CPU pinned at ~200%** (both carriers) for as long as the load runs
- **`agroal_awaiting_count` stays 0** — waiting threads are never accounted as blocked
- the pool grows by **exactly one connection per 5 s**, i.e. one per `post_auth_delay` period. With 300 threads waiting and `max-size=30`, connection creation appears to be **fully serialized**.

## Expected — same run, fast connection creation

Identical burst, `post_auth_delay = 0`:

```
 t(s)   active    await  maxUsed  available     CPU%
    5       30      270       30          0      150
   10       30      270       30          0       28
   ...
   40       30      270       30          0        8
```

The pool reaches `max-size` immediately, surplus threads **park** (`awaiting=270`), CPU drops to single digits. This is the behaviour #203 introduced, and it is correct.

**The only difference between the two runs is connection-creation latency.**

## Profile

async-profiler, 20 s during the spinning run — 4013 samples, **96.16% pass through `handlerFromSharedCache`**:

| share | frame |
|---|---|
| 58.41% | `io.agroal.pool.ConnectionPool.handlerFromSharedCache` |
| 32.89% | `io.agroal.pool.util.StampedCopyOnWriteArrayList.iterator` |
| 4.19% | `clock_gettime` |
| 0.42% | `io.agroal.pool.util.ListenerHelper.fireBeforePoolBlock` |

Representative stack:

```
repro/PoolResource.hit
  io/agroal/pool/DataSource.getConnection
    io/agroal/pool/ConnectionPool.getConnection
      io/agroal/pool/ConnectionPool.internalGetConnection
        io/agroal/pool/ConnectionPool.handlerFromSharedCache
          io/agroal/pool/util/StampedCopyOnWriteArrayList.iterator
```

Raw collapsed stacks are in the reproducer repo.

For contrast, the profile we attached to #201 was 68% `AtomicReferenceFieldUpdater.compareAndSet` + 21% `clock_gettime`. #203 removed that CAS/inline-creation component; the surrounding scan loop remains and now dominates.

That `fireBeforePoolBlock` accounts for only 0.42% is itself telling — the parking path is barely taken.

## Analysis (hypothesis — we measured this rather than read the code path)

When the pool is at `max-size`, `handlerFromSharedCache` knows no new connection can be created and takes the blocking path — hence `awaiting` climbs and CPU is flat. When the pool is below `max-size`, it seems to loop back and rescan `allConnections` while a creation is in flight, rather than waiting for that creation to finish. The longer creation takes, the longer it spins.

On virtual threads this is far worse than on platform threads: a spinning carrier is never released, so with `parallelism` == core count the scheduler stalls and connections being created cannot be handed out promptly — which would also explain the serialization above.

## Impact

For a virtual-thread application, any transient slowdown in connection establishment turns into an unbounded, self-sustaining CPU livelock that outlives the original database problem by hours.

## Workaround

`-Djdk.virtualThreadScheduler.parallelism` set well above the core count. It does not remove the spin; it only ensures the spinning threads cannot monopolise every carrier.

gastaldi:
Fixes #214.

Follow-up to #203 (AG-315). That fix addressed the case where the pool is *at* `max-size`, where threads park correctly. This is a second, distinct spin in the same method that it did not cover.

## Problem

In `ConnectionPool.handlerFromSharedCache`, when the pool is below `max-size` but `acquireCreateConnectionPermit( maxSize )` returns false — other threads already hold every creation permit — the `if` body is skipped entirely and the enclosing `for ( int i = 0; ; i++ )` immediately re-iterates. Nothing on that path waits, and nothing checks the deadline.

So every thread that loses the permit race spins at full speed for as long as the in-flight creations take. The slower connection establishment is, the longer the spin. At `max-size` the `else` branch blocks on `waitAvailableHandler`, which is why the AG-315 scenario looks clean.

The reporter of #214 hit this in production: a managed-PostgreSQL restart made connection establishment slow for a few minutes, after which the application stayed pegged at 100% on both cores for 7 hours and never recovered on its own.

Worth noting this is **not** virtual-thread specific — it reproduces on platform threads just as well. Virtual threads only make it fatal, because a spinning virtual thread never releases its carrier, so with `jdk.virtualThreadScheduler.parallelism` equal to the core count the scheduler stalls entirely.

## Fix

Threads that cannot get a permit now wait on `handlerTransferQueue` instead of spinning, capped at one second and bounded by the acquisition deadline.

A permit is also released when a creation *fails*, and that releases no connection to transfer, so waiters would sleep until the cap. To avoid that, `createAndPoolConnection` hands over a `TRANSFER_RETRY` sentinel on its failure path, waking one waiter immediately — exactly the one that can now take the freed permit. `waitAvailableHandler` returns `null` for the sentinel, so the caller simply re-attempts. The one second cap stays as a backstop for the race where no consumer is waiting yet.

## Results

`PoolGrowthSpinTests` (new) counts acquisition-loop iterations via a listener on `beforePoolBlock`. 16 threads, `max-size=4`, `min-size=0`:

| | loop iterations | CPU | wall clock |
| --- | --- | --- | --- |
| before, slow connect (500ms) | 10,052,300 | 6,099 ms | 753 ms |
| **after**, slow connect (500ms) | **20** | **2 ms** | 716 ms |
| before, fast connect (control) | 603 | 7 ms | 217 ms |
| **after**, fast connect (control) | **16** | **1 ms** | 218 ms |

6.1 seconds of CPU in 753 ms of wall clock, roughly 8 cores, down to nothing. Wall clock is unchanged, so the wait is not costing throughput. The control run is what makes the diagnosis solid: the spin rate is constant and it is establishment latency alone that decides how long it lasts.

## Behaviour change worth reviewing

The fix also makes `acquisitionTimeout` effective on this path, where it was previously ignored — a thread could spin past its timeout indefinitely.

That required updating the second phase of `NewConnectionTimeoutTests.WhenServerAcceptsAndDoNotRespond_ThenAcquisitionTimeoutAndConnectionCreationHangs`, which asserted the old behaviour: with the sole in-flight creation hung inside the driver, a second `getConnection()` used to block indefinitely (*"Which will block the pool, and new connections couldn't be created"*). It now throws `SQLException: Sorry, acquisition timeout!` once the timeout expires. The first `getConnection()` still hangs, since it is stuck in the driver and cannot be interrupted — that part of the test is unchanged.

I believe honouring the configured timeout is the correct contract here, but flagging it explicitly since it is a semantic change rather than a pure optimisation.

## Not addressed here

While investigating I found a second, independent defect, filed as AG-321: since AG-315 sets `collaborate = false` on virtual threads, all creation is dispatched to `housekeepingExecutor`, which is a `ScheduledThreadPoolExecutor` with a core size of 1 and therefore never grows. Connection establishment on virtual threads is fully serialized — this is the "one connection per `post_auth_delay` period" the reporter measured. It needs a separate fix.

## Testing

Full reactor green: 90 tests in `agroal-test`, 88 narayana, 29 spring.


graben (changes requested):
Looks good to me but tests seems flacky between Linux and Windows machines. On my Windows machine:

[ERROR] Failures:
[ERROR]   LifetimeTests.basicLifetimeTest:80 35 connections not created for maxLifetime
[ERROR]   LifetimeTests.minSizeLifetimeTest:160 10 connections not created in time

Which are different to latest run on Github actions where Ubuntu fails withs:

[ERROR] Failures: 
[ERROR]   MSSQLXAReaperRaceTest>XAReaperRaceTestBase.testInterleave:-1 Reaper's rollback() must have completed successfully ==> expected: <true> but was: <false>

I try a respin. But I think there is tuning needed!

EDIT: Got all tests green local and on Github. So fine for me now :)

graben (approved):

gastaldi:
@pavelkryl could you please give this PR a spin and let us know how it goes?

pavelkryl:
Confirmed on the reproducer from #214 — the spin is gone.

**Setup** (unchanged between the two runs, only the Agroal version differs): reproducer at https://github.com/pavelkryl/agroal-slow-connect-repro, Quarkus 3.37.4, JDK 25.0.3, `max-size=30`, `min-size=2`, `acquisition-timeout=10`, `-Djdk.virtualThreadScheduler.parallelism=2`, load `ab -c 300`, each request holds its connection for 0.5 s. Pool starts cold.

Slow connection creation is induced with PostgreSQL's `post_auth_delay = 5s`, which delays only **new** backends and leaves established connections untouched. Verified before each run: a fresh connection takes 5.1 s, an existing one is unaffected.

This branch was built from `refs/pull/215/head` as `3.3-SNAPSHOT` (with JDK 21, for the multi-release sources) and pinned into the reproducer.

| | 3.2.1 | this PR |
|---|---|---|
| process CPU (2 carriers) | **199–207 %** | **5–31 %** |
| `agroal_awaiting_count` | **0** | **287–297** |
| samples in `handlerFromSharedCache` | **96.16 %** (3859 / 4013) | **1.90 %** (4 / 211) |
| total CPU samples in a 20 s profile | 4013 | **211** |
| pool growth | 1 conn / 5 s | 1 conn / 5 s |

async-profiler now shows C2 compiler phases (`PhaseChaitin`, `PhaseLive`, `IndexSetIterator`) at the top — ordinary JIT warm-up. The acquisition loop no longer appears.

Two observations worth recording:

**The metric became truthful.** On 3.2.1 the waiting threads spun while `agroal_awaiting_count` stayed at **0**, so an operator watching pool metrics saw "no contention" while the process pegged every carrier. That zero is what misled our own incident analysis for two days. With this PR the gauge reports the real queue.

**Pool growth rate is unchanged, and that is expected** — one connection per `post_auth_delay` period, since creation is still serialized by the permit. The PR stops the CPU burn while waiting; it does not parallelize creation. Flagging it so the unchanged number isn't read as the fix not taking effect.

For context on why this mattered to us: the production incident behind #214 was a managed-PostgreSQL restart that made connection creation slow for a few minutes. The application then stayed pegged at 100 % on both cores for **7 hours** and never recovered on its own; restarting it relapsed within ~40 seconds, because a cold pool has to grow. Our only workaround was raising `jdk.virtualThreadScheduler.parallelism` well above the core count.

Tested on the JVM only — I have not run this build as a GraalVM native image, which is what we run in production.

Thanks for the quick turnaround on this.

gastaldi:
@barreiro I'd need your review before merging this

graben:
@gastaldi , @barreiro : How does it continue here?

